Issue #2, opened 2016-07-07
Use same style as srobo/website


rgilton:
The style of the pages is currently Jekyll's default.  The style of the docs needs to be similar/identical to that of srobo/website.

PeterJCLaw:
https://github.com/srobo/docs/pull/4 did indeed fix this, though the way the brand is imported has changed a little since then.

PR #4, opened 2016-09-11
Bring docs within style guide

jimmythompson:
This is a first cut at bringing the documentation in line with the style guide , and with the rest of the website, as raised in #2.

screen shot 2016-09-11 at 13 32 53

Some problems:

* the syntax highlighting is still inconsistent,
* there's a bunch of missing images, and
* the sidebar is manually created.


jimmythompson:
There's also a number of visual inconsistencies, like this one:

screen shot 2016-09-11 at 13 39 22

PeterJCLaw:
Why has the IDE folder been renamed here? This doesn't really make sense as "IDE" is an acronym and thus should be capitalised.

PeterJCLaw:
I don't think that we should merge a manually created sidebar as having it will remove the impetus to replace it with an auto-generated one (which is the right solution).

rspanton:
Awesome Jimmy -- thanks for investing some time in this.

PR #36, opened 2016-10-11
Ensure tablet page images are size limited

PeterJCLaw:
This was in the original patch, but appeared unneccesary in the new docs CSS.
It turns out that it is needed after all.

Change-Id: Ia1402a2d8d0496ac80e25a095d298f9e699d8a36

### DIFF
--- a/_sass/docs.scss
+++ b/_sass/docs.scss
@@ -65,6 +65,7 @@
     display: block;
     margin: 0 auto;
     width: initial;
+    max-width: 100%;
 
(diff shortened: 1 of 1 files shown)


richardbarlow:
I've just resized the images to be smaller. But this is a good thing to have anyway to prevent any future mishaps.

PeterJCLaw:
Heh, so much for the high-DPI support which was insisted was needed when
they were first added.

PR #39, opened 2016-10-12
Ensure h2 headings clear floated elements

richardbarlow:
Some of the kit pages have a floated photo of the piece of kit and the
following header should clear the image.

@jimmythompson I'm not sure if this is the correct way to solve this problem. I know the old website was full of clears, but since the new website and the docs have very few clears I suspect that maybe they are not good?


richardbarlow:
This has been fixed in https://github.com/srobo/docs/pull/46

PR #44, opened 2016-11-01
Add a page with kit shipping instructions

richardbarlow:


richardbarlow:
@rspanton could you please sanity check this.

rspanton:
Just read through it.  This seems fine to me.

rspanton:
The only thing I have to add is that it tells them to contact the logistics coordinator if they have a problem with their battery.  Is this also who they should contact if they have a problem with their kit in general?  If so, that's fine.  If not, then this may cause confusion.

team_admin/kit_shipping.md
```diff
@@ -0,0 +1,38 @@
+---
+layout: page
+title: Kit Shipping
+---
+
+# Kit Shipping
+
+This page provides details of how to ship a Student Robotics kit. Please only ship a Student Robotics kit if instructed to do so by the [Kit Logistics Coordinator](mailto:logistics@studentrobotics.org). If you have a kit that you need to ship to us please follow the instructions in the [Packaging](#packaging) section. Once you have your kit packaged, or are looking to get a kit collected from us to be delivered to you, please follow the instructions in the [Arranging a Courier](#arranging-a-courier) section.
```
PeterJCLaw: I would encourage explicitly including the anchor on the target of any internal links so that we're not at the mercy of whatever system happens to be rendering the page.

Note that the style of anchors used here does not match that used elsewhere within our docs (I think we used CamelCase mostly).

team_admin/kit_shipping.md
```diff
@@ -0,0 +1,38 @@
...
+
+## Packaging
+When shipping a complete kit (rather than a single part) it should be shipped in the white 18l Really Useful Box it was provided in. Inside the box, all parts of the kit should either be placed into the jiffy bags that they were provided in or wrapped in bubble wrap. The batteries **must** be placed inside the charging bag provided with the kit. If you suspect that any batteries are damaged then do not ship them and contact <logistics@studentrobotics.org> immediately. Any remaining space in the box should be filled either with paper packaging (old Amazon parcels are a good source of this) or bubble wrap. Please do not use packing peanuts as these are messy and hard to dispose of.
```
PeterJCLaw: Do the teams know that they're meant to keep the packaging it came in? I expect most will have thrown it away unless explicitly told otherwise.

team_admin/kit_shipping.md
```diff
@@ -0,0 +1,38 @@
...
+
+Once all of the parts of the kit are safely packed into the Really Useful Box please tape up the lid and handles as shown in the photo below (to be added).
```
PeterJCLaw: We should either resolve this TODO or just comment it out -- there's little benefit to having a TODO visible on the page.

team_admin/kit_shipping.md
```diff
@@ -0,0 +1,38 @@
...
+
+## Arranging a Courier
+Arranging for a courier to collect the kit/package and deliver it can be done quickly and easily via a range of websites. The sections below outline how this is done for a selection of couriers. For all of these couriers, they will come and collect the package from the collection address you specify.
+
+Since the kits contain lithium polymer batteries there are some restrictions around shipping. We have confirmed that the couriers listed below are happy with shipping packages containing the size and quantity of batteries provided in the kit. Detailed step-by-step instructions for other couriers will be added to this page in due course.
```
PeterJCLaw: This paragraph seems at odds with the information actually provided below -- it suggests that there are multiple couriers (which there aren't) and that step-by-step instructions are missing (yet they seem to be present).

richardbarlow:
Hi Peter,

Thanks for the review. I think I've addressed all of the points you raised. Regarding linking to headers, we are fortunately not at the mercy of an arbitrary system rendering the content since we control what system is used. We are currently using Kramdown (via Github Pages) and the auto ids are an explicit feature and are well defined (http://kramdown.gettalong.org/converter/html.html#auto-ids). If we ever change what system we are using then we will have lots of other things in the repo that are incompatible anyway. Also, the use of camel case elsewhere is a hold-over from the content being imported from the srweb system. Since the new system automatically inserts ids on headers I see no reason to not use them and these are hyphenated rather than camel case.

team_admin/user_accounts.md
```diff
@@ -1,4 +1,5 @@
 ---
+redirect_from: /user_accounts
```
PeterJCLaw: Urgh, shame there's not a way to issue a proper redirect rather than using HTML `<meta>`.

richardbarlow:
> They both had a spec for how they generated the ids, it's just the case that PHP markdown's spec was not to generate them and instead only support you defining your own.

I don't really agree with this argument as it seems to imply that one should not use features of a given piece of software as a previously used piece of software did not have these features. Based upon this argument we wouldn't have any software.

I have however added explicit ids to the headers because this decouples the header text and the links to that header; allowing us to freely change the header text without having to change the links. No automatic system could solve this problem (well, not without preserving some historic state).

Issue #18, opened 2016-10-01
Inline images need sizes


PeterJCLaw:
The majority of the images inline in the docs aren't thumbnails (they probably should be, but that's a lot more work) and so fill the entire page. They should have their size fixed so that they flow with the content.

I'm not sure why this isn't already the case since this worked fine on the old docs and I'm assuming the content was just copied across.

PeterJCLaw:
An example of this is at https://studentrobotics.org/docs/kit/motor_board. It turns out however that my initial interpretation was wrong as that image doesn't have a size specified, but there is CSS which is forcing it to be `width: 100%;` which is clearly undesirable.

PR #46, opened 2016-11-06
Add margins around images to prevent the appearance of overlap

PeterJCLaw:


PeterJCLaw:
Note that this _doesn't_ fix the sizing issue noted in #18.

PR #47, opened 2016-12-23
You must use Chrome Shell on the tablet

richardbarlow:
The tablets have the old pre-chrome Android browser on them and it has
no features and supports nothing. We install 'Chrome Shell' onto the
tablets which does have features and does support things.


PeterJCLaw:
This renders fine for me. `Chrome Shell` sounds like the right name for the app, though I don't have a dev kit to validate that with.

PR #50, opened 2017-01-22
Update brand and styling

jimmythompson:
Updated the brand, and changed some of the styling for the page:

1. Made the background white
2. Use the new header font
3. Added a blue background title, like pages on the PR website
4. Added a sidebar link to the home page, called "Introduction"

header

Added padding round syntax highlighting:

screen shot 2017-01-22 at 23 12 12


PeterJCLaw:
Hi Jimmy,

There's a lot of changes going on in this PR. It suspect would be a lot easier to review (and to understand in the future) if it was broken into sets of changes for each of the major themes highlighted in the PR description. Splitting it up also means that parts of it can be shipped while other parts are still under review/discussion.

I'm guessing that the style updates are essentially a copy/paste from the style and/or website projects (though I wonder whether we should have a better way of transferring the changes, such as submodules for example or at least a common merged history) and so won't have a great deal of history with them.

However, the other changes (such as the "introduction" link [and footer fixes?]) feel much more like things which could easily be done as separate PRs. If those PRs need to depend on this one for some reason, then I think that's fine.

I'm not sure what @rspanton's feelings are for this specific PR, though he has historically requested that large patches be broken into parts for ease of review.

Thanks,
Peter

_includes/footer.html
```diff
@@ -1,32 +1,32 @@
...
-    <div class="d-4-12 m-2-12">
+    <div class="column d-4-12 m-2-12">
       <h4>&nbsp;</h4>
       <ul class="sitemap">
-        <li><a href="/news">News</a></li>
-        <li><a href="/events">Events</a></li>
-        <li><a href="/contact">Contact us</a></li>
+        <li><a href="{{ '/news/' | prepend:site.baseurl }}">News</a></li>
+        <li><a href="{{ '/events/' | prepend:site.baseurl }}">Events</a></li>
+        <li><a href="{{ '/contact/' | prepend:site.baseurl }}">Contact us</a></li>
```
PeterJCLaw: Just noticed this having hit a similar issue on the competition-website pages: prepending this here results in these urls being "...org/docs/contact/" (etc.), which is clearly not desired (since they now 404).

PR #59, opened 2018-10-02
Automatically init the brand submodule if needed

PeterJCLaw:


RealOrangeOne:
I _believe_ `update --init` will also pull the most recent changes in, which is *not* what we want. I may be wrong though?

PeterJCLaw:
`git submodule update --init` will always checkout the version of the submodule which is specified in the parent repo, if that's what you mean?
I agree that that alone would be undesirable, however note that it is configured as the say to create `file '_sass/brand/.git` (I looked up how to do that in the website's `Rakefile`, which uses it to create an SSL certificate). As a result, it should only run if the submodule has not yet been initialised.

That only-if-missing behaviour works for me, though it would be great if you could verify it works for you too.

PR #61, opened 2018-10-21
USB connections go to the ODRIOD

PeterJCLaw:
While the previous generation of power board also contained the brain, the current generation kit has a separate ODRIOD as a brain.


richardbarlow:
The micro usb cable description is similarly wrong as it lists things being connected to the power board, but the power board is now just a peripheral like all of the other boards.

PeterJCLaw:
So it is, I scanned it and saw that the ODROID was mentioned, I should have looked more closely!

PR #64, opened 2018-11-05
We're giving out slightly less of the hook-up wire this year

PeterJCLaw:


PeterJCLaw:
We need to provide less because that's all we have.

Issue #69, opened 2018-11-12
Update the docs for the SR2019 markers


PeterJCLaw:
We need to update both the marker sizes and the constants for each type of marker.

Follows from https://github.com/srobo/tasks/issues/48.

PR #70, opened 2018-11-13
Update the docs for the SR2019 vision spec

PeterJCLaw:
This _doesn't_ include an update to the rules, but does update the archive of the dev markers we'll be using and the mentions of the various marker attributes and constants used by the vision system.

Fixes https://github.com/srobo/docs/issues/69


PeterJCLaw:
Note: I'm not planning to merge this until the corresponding change (https://github.com/PeterJCLaw/sr-robot/pull/1) to `sr.robot` is live.

PR #74, opened 2018-11-18
Remove the inexistant interactive troubleshooter

Adimote:
Turns out the interactive troubleshooter didn't make it over to the
'new' website, we probably shouldn't link to it.


Adimote:
??? it's there now. nvm.

PeterJCLaw:
Huh, I'd thought that this worked. I'd much rather that we fixed it if possible. Is there any reason not to do that?

Adimote:
I think my internet cut out when I was loading the troubleshooter, and jumpt to assumptions it's definitely there.

PR #76, opened 2018-12-03
Fix the docs navbar not displaying

Adimote:
There was previously no implementation for to showing/hiding the navbar on the docs, I've added some JS and CSS animations to emulate that of the srobo/website repo.


PeterJCLaw:
Nice one!

Do you think it would be possible to do this with CSS alone, or do we definitely need the JavaScript?

PeterJCLaw:
A thought for after this merges -- should push this down into the brand submodule? I suspect the other sites which use our brand are also missing this.

PR #78, opened 2019-01-19
Add SVG source files for kit diagrams

kierdavis:
Previously only the rendered PNGs were included in this repo.

These files are copied from the previous home of these diagrams in srweb:
https://github.com/srobo/srweb/tree/f1fe2e8cb271e91dfbe6873d97058f736d11dcae/images/content/kit

There appears to be a discrepancy between the [SVG version of the servo board diagram](https://github.com/kierdavis/docs/blob/dd3634650d3ee715457ef2bf5daa0ecfbc9a83de/images/content/kit/servo_board_v4_diagram.svg) and the [PNG version that's currently live on the docs](https://srobo.github.io/docs/images/content/kit/servo_board_v4_diagram.png), in that the former appears to missing all its text. I haven't looked any further into working out why as it's not my primary goal at the moment.


kierdavis:
The missing text might be related to the use of SVG `flowRegion`s - the servo board diagram is the only diagram that makes use of them (the power board diagram also uses some but they all either contain no text or are a duplicate of an existing label). I'm not familiar enough with either `flowRegion` constructs or with Inkscape to know how to fix this right now, other than by deleting them using a text editor, which will very likely break the diagram in some way.

PeterJCLaw:
Hrm, I'd recommend not editing the files by hand if they're Inkscape generated.

PR #79, opened 2019-01-19
Add documentation about usage of the battery balance lead and about the unused ports on the motor board

kierdavis:
This should hopefully reduce the chance of a team plugging the former
into the latter, as happened a few days ago.

This depends on #78. The changes made on top of those proposed in #78 can be seen [here](https://github.com/kierdavis/docs/compare/kit-diagram-svgs...kierdavis:document-unused-mcv4b-ports?expand=1).


kierdavis:
Huh, not sure how that happened. If I open the SVG back up in Inkscape the arrow is still missing, so I guess my fat fingers must have deleted it while adding the new labels.

I've fixed this, although the `+` and `-` labels are still slightly smaller in the new PNG for me.

side-by-side comparison

kierdavis:
P.S. sorry for rebasing instead of adding new commits; I really need to get out of the habit of doing that.

kierdavis:
The other black rectangle (a 10-pin header) is for programming/debugging the microcontroller, and it is indeed not accessible through the case. I'm tempted to say that we should remove it from the diagram, although it is useful as a landmark (like how the microcontroller and the four big capacitors across the middle are present).

I was using `Inkscape 0.92.3 (2405546, 2018-03-11)` on my NixOS system to generate the diagrams before. I've just tried it using `Inkscape 0.92.1 r15371` on a Debian VM and it appears to have rendered it correctly (the `+` and `-` are the same size as the original. I think it's probably a function of the environment (OS distribution and/or font configuration) rather than the Inkscape version.

PR #80, opened 2019-01-19
Avoid redirects in menu links

PeterJCLaw:
This avoids clients incurring redundant redirects when they load these urls, which is especially useful given that GitHub Pages redirects the canonicalisations in a manner which results in the user ending up on a `srobo.github.io` url rather than our proxy.

This mitigates https://github.com/srobo/website/issues/122.


Adimote:
I've got a fix working in https://github.com/srobo/docs/pull/83 so I'd recommend merging that first

PeterJCLaw:
@kierdavis unfortunately no, unlike Netlify GitHub Pages doesn't have a preview mechanism for branches/PRs.

PeterJCLaw:
@Adimote I've fixed the sidebar issues here, before I saw that you had a fix. I suggest going with this PR and closing #83?

PeterJCLaw:
For posterity, here's the command I used to generate the expected urls: `find . -name '*.md' | grep -Pv '\./(gems|README)' | sed -e 's_\./_- url: /_' -e 's_\.md$__' -e 's_/index$_/_' | sort > _data/sidebar_tree.yaml`

Not immediately shippable as it drops the `title`s, but useful when combined with `git diff -w`. It's also how I found the missing page I added in #81.

PR #81, opened 2019-01-19
Add overlooked link to the IDE user settings page

PeterJCLaw:


PeterJCLaw:
@Adimote I don't think you did, from what I can tell this was always missing!

PR #83, opened 2019-01-19
Change sidebar highlights to allow an appended '/'

Adimote:
Previously if a url in _data/sidebar_tree.yaml had a trailing '/'
character, the sidebar highlighting would not match the page url
and therefore wouldn't highlight the current page.

This change adds some robustness to the matching to strip any
extra '/' characters, allowing us to append them to urls
in the data file again.

This fixes the problems caused in https://github.com/srobo/docs/pull/80


PeterJCLaw:
Ah, I'd looked at fixing this as part of #80. I might copy over the source comment you added though.

PR #92, opened 2019-03-02
1st amendment to the rules

prophile:


PeterJCLaw:
meta: in future, it would be great if we could include a reference to the repo & commit id (or CI link) that the PDF was generated from. This aids greatly in understanding the provenance of binary artefacts.

PR #98, opened 2019-03-10
Explain the purpose of the Motor Board fault LED.

### LINKED ISSUE #96: Explain what Motor Board Fault LED means
The fault LED on the Motor board is just an indicator of bad polarity.

trickeydan:
We had an issue with this at the Hack Day, where the fault LED was lit and we could not work out what it meant. We had to resort to the schematics.

Fixes #96

### DIFF
--- a/kit/motor_board.md
+++ b/kit/motor_board.md
@@ -22,12 +22,13 @@ Board Diagram
 Indicators
 ----------
 
-| LED                    | Meaning                 | Initial power-up state
-|------------------------|-------------------------|------------------------------
-| Power                  | The board is powered    | On
-| M{0,1} Speed/Direction | Brightness indicates speed, colour indicates direction | Off
-| USB Power              | The USB interface is powered | On
-| USB Data               | Data is being transferred to/from the board | Off
+| LED                    | Colour     | Meaning                                                | Initial power-up state |
+|------------------------|------------|--------------------------------------------------------|------------------------|
+| Power                  | Green      | The board is powered                                   | On                     |
+| Power                  | Red        | If lit, the polarity is wrong for Power In             | Off                    |
+| M{0,1} Speed/Direction | Red / Blue | Brightness indicates speed, colour indicates direction | Off                    |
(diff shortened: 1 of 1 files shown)


PeterJCLaw:
@trickeydan not sure if this PR was ready for review yet, but I can't see any actual change here. There seems to be some reformatting of the LEDs table, but that's it. Am I missing something?

trickeydan:
> @trickeydan not sure if this PR was ready for review yet, but I can't see any actual change here. There seems to be some reformatting of the LEDs table, but that's it. Am I missing something?

I pasted the wrong table!

trickeydan:
Correct table now force-pushed to remove commit spam

kit/motor_board.md
```diff
@@ -22,12 +22,13 @@ Board Diagram
...
-| LED                    | Meaning                 | Initial power-up state
-|------------------------|-------------------------|------------------------------
-| Power                  | The board is powered    | On
-| M{0,1} Speed/Direction | Brightness indicates speed, colour indicates direction | Off
-| USB Power              | The USB interface is powered | On
-| USB Data               | Data is being transferred to/from the board | Off
+| LED                    | Colour     | Meaning                                                | Initial power-up state |
+|------------------------|------------|--------------------------------------------------------|------------------------|
+| Power                  | Green      | The board is powered                                   | On                     |
+| Power                  | Red        | If lit, the polarity is wrong for Power In             | Off                    |
```
RealOrangeOne: To match the pattern of the below line, we should drop these to 1 line.

trickeydan: The other LEDs do not have multiple meaningful colours.

PeterJCLaw: There's an existing precedent for the multi-line approach in the [power board docs](https://studentrobotics.org/docs/kit/power_board#indicators), it would be great if we could use that.

RealOrangeOne: > The other LEDs do not have multiple meaningful colours.

The line below my comment does exactly this. We should be consistent within the same table

PR #99, opened 2019-03-10
Document buffer overflow bug on Power Board

### LINKED ISSUE #97: Document 32 buzz buffer limit
Competitors are getting USB errors and couldn't understand why. This is due to the ring buffer on the power board overflowing.

trickeydan:
When the sample ring buffer in [piezo.c](https://www.studentrobotics.org/cgit/boards/power-v4-fw.git/tree/piezo.c#n22) overflows, the code will crash and cause a reset of the board.

Teams have come across this and were asking about it at the Tech Day.

Closes #97

### DIFF
--- a/kit/power_board.md
+++ b/kit/power_board.md
@@ -87,6 +87,11 @@ The case measures 83x99x24mm. Don't forget that the cables will stick out.
 
 [^2]: If overall current limit is exceeded, the Power Board will turn off and start beeping.
 
+Notes
+-----
+
+There is an internal ring buffer that stores a queue of buzzes for the Piezo buzzer. If the number of buzzes queued exceeds 32, the buffer may overflow and cause your code to crash. This is a known issue with the board.
+
 
 Designs
 -------


kierdavis:
Meta: when I click the cgit link in the issue description, I'm actually taken to a completely different page on a different website.

kierdavis:
Has an issue been opened for fixing this in the firmware? (that is, returning a more useful error response than a generic USB stall)

RealOrangeOne:
> Meta: when I click the cgit link in the issue description, I'm actually taken to a completely different page on a different website.

FTFY

trickeydan:
> Has an issue been opened for fixing this in the firmware? (that is, returning a more useful error response than a generic USB stall)

It has been in the `j5` version of the firmware, which to my knowledge is the copy most likely to be worked on in the future. SR don't have a copy of the repo on GitHub.

PeterJCLaw:
I've reported this against the (newly added) GitHub clone of the SR power board firmware: https://github.com/srobo/power-v4-fw/issues/3.

trickeydan:
I'm going to put this PR on hold pending potential updates to the firmware that will either fix this, or return a reasonable error message. The primary issue is that it crashes the USB bus.

kierdavis:
@trickeydan whilst I agree that a firmware update is the best solution in the long term, I don't think it's guaranteed to happen for the next competition cycle. I'd be in favour of merging this PR now and reverting it later once the underlying issue is fixed.

trickeydan:
Would be good to see this sorted before kickstart

kierdavis:
A firmware update is not happening for SR2020 because the boards have now been through pre-competition testing.

As @RealOrangeOne requested, I think this should in the API section, near where we talk about the `beep` method.

trickeydan:
This issue is no longer a problem with the new firmware. There is no ring buffer.

PR #108, opened 2019-10-07
Updates for sr2020 kickstart

TomWheal:
This is what the kit will contain. 

The change in wire is because I witnessed a lot of teams using .5mm wire for power. So by providing thick(1mm) power wire and thin (.5mm) signal wire, it makes it a lot easier for them. Also by providing more it makes it less likely they'll use the signal wire.

Also new USBs all round.

### DIFF
--- a/kit/index.md
+++ b/kit/index.md
@@ -37,16 +37,13 @@ As well as the aforementioned boards the kit also contains the following items.
 | Battery Charger        | 1   | iMAX B6 or HobbyKing E4                                               | N/A                                         | Documentation is available on [charging the batteries](/docs/kit/batteries) using this
 | USB Hub                | 2   | 7-Port                                                                | [StarTech ST7202USBGB][ST7202USBGB]         | The hub does *not* require a power cable
 | USB WiFi Adapter       | 2   | 300Mbps Mini Wireless N USB Adapter                                   | [TP-Link TL-WN823N][TL-WN823N]              |
-| USB Memory Stick       | 1   | Kingston Datatraveler G2 2GB or G3 8GB                                | N/A (discontinued)                          |
+| USB Memory Stick       | 1   | Kingston Datatraveler 16Gb SE9H                                       | [DTSE9H/16GB][DTSE9H-16GB]                         |
 | Standard USB cable     | 3   | Standard USB A to B connector                                         | N/A                                         | For connecting the ruggeduino (via a hub) and USB hubs to the ODROID
 | Micro USB cable        | 5   | Standard USB A to Micro USB B                                         | N/A                                         | For connecting the motor boards, servo boards and power board to the ODROID
-| Red Hook-up Wire       | 1   | 2.5m<br />0.5mm<sup>2</sup><br />4A                                     | [RS 361-614][RS-361-614]                    | General purpose wire
-| Black Hook-up Wire     | 1   | 2.5m<br />0.5mm<sup>2</sup><br />4A                                     | [RS 361-579][RS-361-579]                    | General purpose wire
-| Blue Hook-up Wire      | 1   | 2.5m<br />0.5mm<sup>2</sup><br />4A                                     | [RS 361-585][RS-361-585]                    | General purpose wire
-| Yellow Hook-up Wire    | 1   | 2.5m<br />0.5mm<sup>2</sup><br />4A                                     | [RS 361-636][RS-361-636]                    | General purpose wire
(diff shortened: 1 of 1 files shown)


kit/index.md
```diff
@@ -37,16 +37,13 @@ As well as the aforementioned boards the kit also contains the following items.
...
-| Black Power Wire       | 1   | 1.8m<br />1mm<sup>2</sup><br />10A                                    | [RS 361-715][RS-361-715]                    | To connect power from the power board to the motor/servo boards
+| Blue Hook-up Wire      | 1   | 2.5m<br />0.5mm<sup>2</sup><br />4A                                   | [RS 361-585][RS-361-585]                    | General purpose wire
+| Yellow Hook-up Wire    | 1   | 2.5m<br />0.5mm<sup>2</sup><br />4A                                   | [RS 361-636][RS-361-636]                    | General purpose wire
+| Red Power Wire         | 1   | 5m<br />1mm<sup>2</sup><br />10A                                      | [RS 361-759][RS-361-759]                    | To connect power from the power board to the motor/servo boards
+| Black Power Wire       | 1   | 5m<br />1mm<sup>2</sup><br />10A                                      | [RS 361-715][RS-361-715]                    | To connect power from the power board to the motor/servo boards
```
PeterJCLaw: We normally (fairly strongly) encourage teams to use colours other than red or black for connecting from the motor boards to their motors, in order that those lines not appear to have a ground.
(The [rules _require_](https://github.com/srobo/rules/blob/master/regs.tex#L54-L57) that black wire be used for all ground connections and for nothing else.)
This is the primary reason for the inclusion of the blue power wire.
What's the reasoning for removing the blue power wire?

(:+1: to only providing each colour of wire in a single thickness though)

kit/index.md
```diff
@@ -37,16 +37,13 @@ As well as the aforementioned boards the kit also contains the following items.
 | Battery Charger        | 1   | iMAX B6 or HobbyKing E4                                               | N/A                                         | Documentation is available on [charging the batteries](/docs/kit/batteries) using this
 | USB Hub                | 2   | 7-Port                                                                | [StarTech ST7202USBGB][ST7202USBGB]         | The hub does *not* require a power cable
 | USB WiFi Adapter       | 2   | 300Mbps Mini Wireless N USB Adapter                                   | [TP-Link TL-WN823N][TL-WN823N]              |
-| USB Memory Stick       | 1   | Kingston Datatraveler G2 2GB or G3 8GB                                | N/A (discontinued)                          |
+| USB Memory Stick       | 1   | Kingston Datatraveler 16Gb SE9H                                       | [DTSE9H/16GB][DTSE9H-16GB]                         |
```
PeterJCLaw: :tada:

kit/index.md
TomWheal: > What's the reasoning for removing the blue power wire?
I completely missed it..... However, you make a very good point. I am concerned with the number of teams that wire motors backwords. Ill order 2.5mm of 1mm blue wire for the kit, however in the future we should look at maybe some blue with red/black striped wire.

TomWheal: > What's the reasoning for removing the blue power wire?

I missed it tbh... Thinking of replacing this with 2.5m of green 0.75mm wire. Thoughts? We don't need the same thickness as the power wire, and it would be good to ensure competitors aren't wiring up motors using signal wires.

PeterJCLaw: Given that that the motors are essentially the most power hungry component in the kit (and a single one is able to consume the entire power output of a motor board), I think it's reasonable for the wire motor-side to be the same as that power-side of a motor board.

I don't feel strongly about the colour, but would note that green (or green & yellow) is often used to mean "earth" which I suspect may have influenced the previous choice of blue.

RealOrangeOne:
These changes look pretty good to me!

Once this is merged, it'll also need updating in the rules (branching off https://github.com/srobo/rules/pull/30)!

kit/index.md
rgilton: Yes, please don't get rid of the non red/black thicker wire.  Don't care what colour it is as long as it's not red/black.  Required for motor connections.

Also, 5m of power cabling is quite wasteful (10 robot widths!) -- most of this will end up in the bin, so not a good use of SR's money nor the planet!  The 1.8m from before was fine.

TomWheal: I agree in retrospect, I got out 5m of wire in labs today, I think 1.8 might be pushing it. but 2.5m is plenty. Also a nice round number for 40 kits form 100m of wire

TomWheal:
> Once this is merged, it'll also need updating in the rules (branching off [srobo/rules#30](https://github.com/srobo/rules/pull/30))!

We don't ask for wire back so this isn't actually required the core kit isnt changing.

Issue #109, opened 2019-10-23
Favicon on docs not showing


TomWheal:
The favicon seems to be missing on the docs pages

PR #112, opened 2019-10-23
Correct favicon

RealOrangeOne:
The frontend is looking for a file at `/mages/favicon.ico`, so we should actually put a file there!

We could change it to just link to the favicon from the website, but that feels a little strange to me!

Fixes https://github.com/srobo/docs/issues/109


PeterJCLaw:
> We could change it to just link to the favicon from the website, but that feels a little strange to me!

I think this would be preferable to avoid them getting out of sync. An alternative would be to have the favicon come from the brand submodule directly.

IMO having a duplicate file in each subsite repo (we have at least four) is not sustainable.

PR #122, opened 2020-05-15
Add basic page for competition simulator

RealOrangeOne:
Fixes https://github.com/srobo/competition-simulator/issues/12
Fixes https://github.com/srobo/competition-simulator/issues/42
Fixes https://github.com/srobo/competition-simulator/issues/55

Add a page for the [competition simulator](https://github.com/srobo/competition-simulator/)

Heavily based on https://github.com/srobo/competition-simulator/pull/31/

The download link here is direct to the repo, so it will need open-sourcing before this page is pushed live.


competition-simulator.md
```diff
@@ -0,0 +1,57 @@
...
+title: Competition Simulator
+---
+
+Competition Simulator
+============
+
+For SR2020, a virtual competition will be run using a brand new simulator.
+
+<div class="info">
+  This is not the same as the <a href="/docs/programming/simulator/">regular simulator</a>.
```
PeterJCLaw, line 12: Suggestion: invert this negative:
```suggestion
  This is different to the <a href="/docs/programming/simulator/">regular simulator</a>.
```

competition-simulator.md
```diff
@@ -0,0 +1,57 @@
...
+</div>
+
+## Installation
+
+### Prerequisites
+
+This simulator is based off [Webots](https://cyberbotics.com/#download), which will need to be downloaded and installed.
+
+You will also need Python installed. The simulator supports >=3.4.
```
PeterJCLaw, line 21: This conflicts with the dev instructions in https://github.com/srobo/competition-simulator/pull/31, which indicates that this is Python 2.7.

competition-simulator.md
```diff
@@ -0,0 +1,57 @@
...
+
+- [Camera Controls](https://www.cyberbotics.com/doc/guide/the-3d-window#navigation-in-the-scene)
+- [Graphics settings](https://www.cyberbotics.com/doc/guide/preferences#opengl) (Useful if Webots is running slowly)
+
+## Developing your code
+
+The code for your robot is `controllers/example_controller/example_controller.py`. An example skeleton code is provided for convenience.
+
+<div class="warning">
+  Please refrain from editing any code in the <code>sr</code> module directory. This contains the code which maps the SR API onto the webots API. Changes to this will not be carried forward into a competition environment.
```
PeterJCLaw, line 50: > Changes to this will not be carried forward into a competition environment.

This feels like it could be misread, to imply something about the façade not being present in the competition. Perhaps something like
```suggestion
  Please refrain from editing any code in the <code>sr</code> module directory. This contains the code which maps the SR API onto the webots API.
  Any changes you make to this directory will not be present in the competition environment.
```

competition-simulator.md
```diff
@@ -0,0 +1,57 @@
...
+
+## Installation
+
+### Prerequisites
+
+This simulator is based off [Webots](https://cyberbotics.com/#download), which will need to be downloaded and installed.
+
+You will also need Python installed. The simulator supports >=3.4.
+
+### Installing the simulator
```
PeterJCLaw, line 23: Should we use a different term than "simulator" here? As I understand it, Webots is doing the actual simulation -- the extra files we're providing are more a description of the world.

In particular it reads oddly to me that below we specify "Download the simulator" but then instruct to use the Webots IDE to open that simulator.

I think it's because to me a "simulator" is something which you'd expect to run, rather than work with.

Even just "simulation" might be a clearer framing as that's perhaps clearer that it's a thing you're going to work with.

competition-simulator.md
```diff
@@ -0,0 +1,57 @@
...
+
+1. [Download the simulator](https://github.com/srobo/competition-simulator/archive/master.zip), and unzip it somewhere on your computer.
```
PeterJCLaw, line 25: I don't know what the merge policy of the target repo is, however it is a little unusual that this references `master` (a floating branch name) rather than a fixed commit hash or a tag.

competition-simulator.md
antoinepetty: Fixed the linked guide accordingly. We've opted to go with Python 3 because Python 2 seems quite buggy in WeBots

competition-simulator.md
antoinepetty: I like the idea of using the word "simulation" instead of "simulator" to refer to our project in WeBots

competition-simulator.md
```diff
@@ -0,0 +1,57 @@
...
+
+<div class="info">
+  This is not the same as the <a href="/docs/programming/simulator/">regular simulator</a>.
+</div>
+
+## Installation
+
+### Prerequisites
+
+This simulator is based off [Webots](https://cyberbotics.com/#download), which will need to be downloaded and installed.
```
antoinepetty, line 19:
```suggestion
Our simulator is based in [Webots](https://cyberbotics.com/#download), which will need to be downloaded and installed.
```

competition-simulator.md
```diff
@@ -0,0 +1,57 @@
...
+**Important:** Changes to the world must happen with the simulation paused at 0:00. If e.g. you move an object at a different time, rewinding back to the start will delete your changes.
+
+### Useful links
+
+- [Camera Controls](https://www.cyberbotics.com/doc/guide/the-3d-window#navigation-in-the-scene)
+- [Graphics settings](https://www.cyberbotics.com/doc/guide/preferences#opengl) (Useful if Webots is running slowly)
+
+## Developing your code
+
+The code for your robot is `controllers/example_controller/example_controller.py`. An example skeleton code is provided for convenience.
```
antoinepetty, line 47:
```suggestion
The code for your robot is `controllers/sr_controller/sr_controller.py`. An example skeleton code is provided for convenience.
```
Because of https://github.com/srobo/competition-simulator/pull/35

competition-simulator.md
RealOrangeOne: This is intentional, so that it's always pointing to the latest version.

We could use a regular tagging-based release if that's better?

antoinepetty: I think it's best to keep some sort of formal release process externally so it's explicit which version of the API the team is downloading at a given time (and in case a big bug makes it into Master at some point). Github has a "release" management system but regular git tags works for me too

RealOrangeOne: I think using GitHub releases (which back onto tags) is probably fine for this. I'll leave this link unpopulated until we've got said release link

competition-simulator.md
RealOrangeOne: Have addressed in the places it makes sense

competition-simulator.md
```diff
@@ -0,0 +1,120 @@
...
+- Front Left
+- Front Right
+- Left
+- Right
+- Back Left
+- Back Right
+
+The `analogue_read` method will return the distance in centimetres, however only measure up to 30cm.
+
+### Camera
```
PeterJCLaw, line 99: This will need considerable updates in light of the more complete API presented by https://github.com/srobo/competition-simulator/pull/33 now.

competition-simulator.md
```diff
@@ -0,0 +1,120 @@
...
+Analogous to ultrasound sensors, distance sensors allow you to retrieve the distance between your robot and an object.
+
+- Front Left
+- Front Right
+- Left
+- Right
+- Back Left
+- Back Right
+
+The `analogue_read` method will return the distance in centimetres, however only measure up to 30cm.
```
PeterJCLaw, line 97: Why only up to 30cm? This seems somewhat arbitrary. I had thought that ultrasound responded up to several metres away?

competition-simulator.md
```diff
@@ -0,0 +1,120 @@
...
+- Back
+- Between gripper fingers
+- Left gripper finger
+- Right gripper finger
+
+Using the `digital_read`  method, you'll receive a `bool` telling you whether the switch is current actuated.
+
+#### Distance Sensors
+
+Analogous to ultrasound sensors, distance sensors allow you to retrieve the distance between your robot and an object.
```
PeterJCLaw, line 88: Where are these attached? That is: which pins should competitors "use" to get these data?

competition-simulator.md
```diff
@@ -0,0 +1,120 @@
...
+  This is different to the <a href="/docs/programming/simulator/">regular simulator</a>.
+</div>
+
+## Installation
+
+### Prerequisites
+
+This simulation is based in [Webots](https://cyberbotics.com/#download), which will need to be downloaded and installed.
+
+You will also need Python installed. The simulator supports >=3.4.
```
PeterJCLaw, line 21: ❌ This has changed, we're requiring 3.5 now.

competition-simulator.md
```diff
@@ -0,0 +1,120 @@
...
+
+### Installing the simulation
+
+1. [Download the simulation](), and unzip it somewhere on your computer.
```
PeterJCLaw, line 25: ❌ Incomplete link (I realise this is intentional while we haven't done a first release, highlighting so we don't miss this).

competition-simulator.md
```diff
@@ -0,0 +1,120 @@
...
+- [Graphics settings](https://www.cyberbotics.com/doc/guide/preferences#opengl) (Useful if Webots is running slowly)
+
+## Developing your code
+
+The code for your robot is `controllers/sr_controller/sr_controller.py`. An example skeleton code is provided for convenience.
+
+<div class="warning">
+  Please refrain from editing any code in the <code>sr</code> module directory. This contains the code which maps the SR API onto the webots API.
+  Any changes you make to this directory will not be present in the competition environment.
+</div>
```
PeterJCLaw, line 52: In the preceding section we appear to encourage the competitors to also modify the world state. Should we (either there or here) also clarify that world state changes won't be present in the real competition?

competition-simulator.md
```diff
@@ -0,0 +1,120 @@
...
+
+### Updates
+
+Occasionally, we may release an update to the simulation. To update, you will need to copy your code our of of the `competition-simulator-master` directory, delete the directory, and re-download it using the above link. Then you can replace the new `controllers/sr_controller/sr_controller.py` with your existing code.
```
PeterJCLaw, line 56: Hrm, this approach feels quite likely to result in "oops I accidentally deleted our code" from people just blatting the new directory on top of the old one.

Could we do something to avoid that? For example, can we arrange it so that the competitors code is meant to live _outside_ the directory they downloaded?

Related to this: does Webots come with a Python IDE, or are they going to use something of their own? If the former, would moving the code outside this structure complicate that?

competition-simulator.md
```diff
@@ -0,0 +1,120 @@
...
+
+## Programming Interface
+
+**WIP**
+
+Unless otherwise stated, the simulator’s API is the same as the real SR API described in the [docs]({{ site.baseurl }}/docs/programming/).
```
PeterJCLaw, line 62:
```suggestion
Unless otherwise stated, the simulator’s API is the same as the real SR API described in the [programming docs]({{ site.baseurl }}/docs/programming/).
```

competition-simulator.md
```diff
@@ -0,0 +1,120 @@
...
+<div class="warning">
+  Please refrain from editing any code in the <code>sr</code> module directory. This contains the code which maps the SR API onto the webots API.
+  Any changes you make to this directory will not be present in the competition environment.
+</div>
+
+### Updates
+
+Occasionally, we may release an update to the simulation. To update, you will need to copy your code our of of the `competition-simulator-master` directory, delete the directory, and re-download it using the above link. Then you can replace the new `controllers/sr_controller/sr_controller.py` with your existing code.
+
+## Programming Interface
```
PeterJCLaw, line 58: (for lack of a better place to put this): Can competitors submit their own robot designs?
If not, perhaps we should introduce the robot first, with a picture, to give an idea of what it can do, before introducing how to use it?

competition-simulator.md
```diff
@@ -0,0 +1,120 @@
...
+### Prerequisites
+
+This simulation is based in [Webots](https://cyberbotics.com/#download), which will need to be downloaded and installed.
+
+You will also need Python installed. The simulator supports >=3.4.
+
+### Installing the simulation
+
+1. [Download the simulation](), and unzip it somewhere on your computer.
+2. Using the Webots IDE, open the `worlds/Arena.wbt` file
```
PeterJCLaw, line 26: ❌ We also need to tell people that they may need to configure Webots to use Python 3 if they have more than one Python installed.

PeterJCLaw, line 26: ❌ We should tell people to ignore the scary warning about your graphics card not being good enough.

competition-simulator.md
```diff
@@ -0,0 +1,124 @@
...
+  This is different to the <a href="/docs/programming/simulator/">regular simulator</a>.
+</div>
+
+## Installation
+
+### Prerequisites
+
+This simulation is based in [Webots](https://cyberbotics.com/#download), which will need to be downloaded and installed.
+
+You will also need Python installed. The simulator supports >=3.5.
```
PeterJCLaw, line 21: We should clarify that we're not providing any additional libraries.

PeterJCLaw, line 21: We should specify what version of Python the competition will be run on.

competition-simulator.md
RealOrangeOne: Ultrasound does (although accuracy drops off a cliff after a few metres), although I remember @antoinepetty mentioning something about this maximum range

competition-simulator.md
RealOrangeOne: I've tracked this in the PR description :+1:

competition-simulator.md
RealOrangeOne: Webots does have an editor, and yes moving the code outside would complicate things. I've intentionally written things such that they can use their own editors.

Moving the code primarily outside would be possible, and we could probably do that with a custom controller.

Alternatively, we should encourage regular backups of code, and good version control practice, regardless of this update mechanism

competition-simulator.md
RealOrangeOne: Teams cannot submit their own robot designs.

I was hoping that the API documentation does enough for what the robot can do?

competition-simulator.md
RealOrangeOne: I believe the "Changing your version of Python" covers this?

competition-simulator.md
RealOrangeOne: I'm not sure we can commit to that quite yet.

competition-simulator.md
antoinepetty: I'm tempted to suggest they fork the git repo for their development. (Although this doesn't seem like the perfect solution)

RealOrangeOne: That's definitely not a viable option, as many competitors won't have the slightest idea what `git` is, let alone how to use it properly.

competition-simulator.md
antoinepetty: Yeah as far as I could tell from the webots docs it's a limit on their end.

competition-simulator.md
antoinepetty: The latest commit has this detail so I'll set this as resolved

competition-simulator.md
antoinepetty: I think having a picture perhaps with labels is a good idea. I'm happy to take this job if you want

competition-simulator.md
antoinepetty: This is done so I'll mark as resolved

competition-simulator.md
antoinepetty: Fixed in latest commit so will mark as resolved

competition-simulator.md
antoinepetty: This doesn't exist anymore so will mark as resolved

antoinepetty:
Some final things before I approve:

- Need to explicitly mention Two Colours as the game
- Need to update the vision section
- Could do with a picture of the robot and arena mainly for aesthetics :)

competition-simulator.md
```diff
@@ -0,0 +1,120 @@
...
+
+Within the IDE, there are a few different panels:
+
+- In the centre of your screen is the 3D simulated view of the arena
+- On the left is a tree hierarchy of all elements in this "world"
+- On the right is your text editor for modifying the `sr_controller` code for example
+- At the bottom is your console
+- At the top are your general controls which include the time controls. Press the centre play button to run the simulation at normal speed.
+
+**Important:** Changes to the world must happen with the simulation paused at 0:00. If e.g. you move an object at a different time, rewinding back to the start will delete your changes.
```
antoinepetty, line 46: Oh I thought we were removing this part? If not I think Peter's comment about us stating explicitly that we're not including world changes is valid

competition-simulator.md
```diff
@@ -0,0 +1,120 @@
...
+
+Occasionally, we may release an update to the simulation. To update, you will need to copy your code out of the `competition-simulator-master` directory, delete the directory, and re-download it using the above link. Then you can replace the new `controllers/sr_controller/sr_controller.py` with your existing code.
+
+## Robot
+
+![]({{ site.baseurl }}/resources/competition-simulator/robot-front.png)
+
+## Programming Interface
+
+**WIP**
```
antoinepetty, line 72: Remove this label

competition-simulator.md
```diff
@@ -0,0 +1,120 @@
...
+The code for your robot is `controllers/sr_controller/sr_controller.py`. An example skeleton code is provided for convenience.
+
+<div class="warning">
+  Please refrain from editing any code in the <code>sr</code> module directory. This contains the code which maps the SR API onto the webots API.
+  Only your controller code will be present in the competition environment.
+</div>
+
+### Updates
+
+Occasionally, we may release an update to the simulation. To update, you will need to copy your code out of the `competition-simulator-master` directory, delete the directory, and re-download it using the above link. Then you can replace the new `controllers/sr_controller/sr_controller.py` with your existing code.
```
antoinepetty, line 64: Do we want to recommend version control systems so teams can share code?

competition-simulator.md
```diff
@@ -0,0 +1,120 @@
...
+
+### Ruggeduino
+
+Your robot has 5 microswitches and 6 distance sensors, attached to the digital and analogue pins respectively. These are all attached to a single ruggeduino.
+
+Because these sensors are pre-attached to the ruggeduino, you do not need to set its `pin_mode`.
+
+#### Microswitches
+
+The microswitches are attached to digital pins 0-4:
```
antoinepetty, line 88: They're actually on pins 2-6

competition-simulator.md
```diff
@@ -0,0 +1,120 @@
...
+Analogous to ultrasound sensors, distance sensors allow you to retrieve the distance between your robot and an object. These are attached to analogue pins 0-5:
+
+- Front Left
+- Front Right
+- Left
+- Right
+- Back Left
+- Back Right
+
+These are shown as blue coloured blocks on the robot. The `analogue_read` method will return the distance in centimetres, however only measure up to 30cm.
```
antoinepetty, line 109: These are now in metres I think https://github.com/srobo/competition-simulator/pull/71

competition-simulator.md
```diff
@@ -0,0 +1,120 @@
...
+
+### Camera
+
+Your robot has a camera, which is attached to the top of your gripper. A live preview of what the camera sees is shown in the top-left corner. The vision system uses Webots' object recognition, rather than [fiducial markers]({{ site.baseurl }}/programming/sr/vision/markers/).
```
antoinepetty, line 113:
```suggestion
Your robot has a camera, which is attached to the top of your robot. A live preview of what the camera sees is shown in the top-left corner. The vision system uses Webots' object recognition, rather than [fiducial markers]({{ site.baseurl }}/programming/sr/vision/markers/).
```

competition-simulator.md
RealOrangeOne: I'll put in a warning about backing up code

competition-simulator.md
PeterJCLaw: Them using forks makes updates much harder (merge conflicts also being possible) as well as meaning that we'd effectively need to support every commit, rather than a small number of blessed releases.

PeterJCLaw: @RealOrangeOne I think we should consider this blocking for release -- I don't think we're going to get a second chance to change how teams set up their development environments.

I suggest we fix this by having our `sr_controller.py` look for their code in some known location, perhaps a sibling `robot` directory to the place the simulator is checked out to.

From my testing, the built-in editor lets you open files anywhere you like and didn't seem to have any special affinity for the files within itself anyway.

If we're going to encourage that they use other editors (which I think we should - that editor is pretty bad), then we might as well have the code somewhere they can feel more in control of it. There might be a bit more setup for their editors, but I think that that's worth the trade-off of not losing code (especially since the IDE made it _hard_ to lose code, but this current mechanism makes it a loaded foot-gun).

competition-simulator.md
```diff
@@ -0,0 +1,131 @@
...
+### Installing the simulation
+
+1. [Download the simulation](https://github.com/srobo/competition-simulator/archive/0.1.0.zip), and unzip it somewhere on your computer.
+2. Using the Webots IDE, open the `worlds/Arena.wbt` file.
+
+You may receive a warning about your computer's GPU not being good enough, which can be ignored.
+
+#### Changing your version of Python
+
+If webots is picking up the incorrect version of Python, you'll need to change it. This can be done using `Tools > Preferences > General > Python command`. You'll need to ensure a matching version of Python is installed.
```
PeterJCLaw, line 34: I think we worked out that we can avoid this by ensuring there's a suitable shebang in the controller file?

competition-simulator.md
```diff
@@ -0,0 +1,131 @@
...
+
+## Overview
+
+Within the IDE, there are a few different panels:
```
PeterJCLaw, line 38:
```suggestion
Within the Webots IDE, there are a few different panels:
```

competition-simulator.md
```diff
@@ -0,0 +1,131 @@
...
+
+- In the centre of your screen is the 3D simulated view of the arena
+- On the left is a tree hierarchy of all elements in this "world"
+- On the right is your text editor for modifying the `sr_controller` code for example
```
PeterJCLaw, line 42: Is `sr_controller` the name we're going for? I think we're now encouraging them to use a `robot.py` file?

competition-simulator.md
```diff
@@ -0,0 +1,131 @@
...
+  Only your controller code will be present in the competition environment.
+</div>
+
+<div class="warning">
+  You will be responsible for backing up and versioning your code, and collaborating with your fellow team members. The <a href="https://studentrobotics.org/ide/">IDE</a> is unsuitable for this simulator.
+</div>
+
+### Updates
+
+Occasionally, we may release an update to the simulation. To update, you will need to delete the `competition-simulator-master` directory, and re-download it using the above link.
```
PeterJCLaw, line 77: This directory name is different to the one mentioned in the tree above -- is that intentional?

competition-simulator.md
```diff
@@ -0,0 +1,131 @@
...
+
+#### Microswitches
+
+The microswitches are attached to digital pins 2-6:
+
+- Front
+- Back
+- Between gripper fingers
+- Left gripper finger
+- Right gripper finger
```
sedders123, line 105: I think we should explicitly mention which pin each of these switches  belong to.
```suggestion
- Front - Pin  2
- Back - Pin 3
- Between gripper fingers - Pin 4
- Left gripper finger - Pin 5
- Right gripper finger - Pin 6
```

competition-simulator.md
```diff
@@ -0,0 +1,131 @@
...
+  You will be responsible for backing up and versioning your code, and collaborating with your fellow team members. The <a href="https://studentrobotics.org/ide/">IDE</a> is unsuitable for this simulator.
+</div>
+
+### Updates
+
+Occasionally, we may release an update to the simulation. To update, you will need to delete the `competition-simulator-master` directory, and re-download it using the above link.
+
+## Robot
+
+![]({{ site.baseurl }}/resources/competition-simulator/robot-front.png)
```
PeterJCLaw, line 81: Should we say something about the robot?

Things I think we should mention here:
- that this is a pre-built shell, which supports our API
- that building a robot model is complicated and we expect teams to use the pre-built shell
- maybe some comments about what it can do?

competition-simulator.md
```diff
@@ -0,0 +1,131 @@
...
+#### Distance Sensors
+
+Analogous to ultrasound sensors, distance sensors allow you to retrieve the distance between your robot and an object. These are attached to analogue pins 0-5:
+
+- Front Left
+- Front Right
+- Left
+- Right
+- Back Left
+- Back Right
```
sedders123, line 118: Similarly I feel being explicit here would be useful
```suggestion
- Front Left - Pin 0
- Front Right - Pin 1
- Left - Pin 2
- Right - Pin 3
- Back Left - Pin 4
- Back Right - Pin 5
```

competition-simulator.md
```diff
@@ -0,0 +1,131 @@
...
+
+![]({{ site.baseurl }}/resources/competition-simulator/robot-front.png)
+
+## Programming Interface
+
+Unless otherwise stated, the simulator’s API is the same as the real SR API described in the [programming docs]({{ site.baseurl }}/programming/).
+
+### Motors
+
+Your robot has 2 motor boards attached, each with 2 motors. Board `0` has the left wheel in port `m0`, and the right wheel in `m1`. Board `1` has the gripper lift motor in `m0`, and the finger motors in `m1`.
```
PeterJCLaw, line 89: Should we note explicitly that the boards do not have part codes, so that the `R.motors['SR0ABC']` spelling won't work?

competition-simulator.md
```diff
@@ -0,0 +1,131 @@
...
+
+These are shown as blue coloured blocks on the robot. The `analogue_read` method will return the distance in metres, however only measure up to 30cm.
+
+### Camera
+
+Your robot has a camera, which is attached to the top of your robot. A live preview of what the camera sees is shown in the top-left corner. The vision system uses Webots' object recognition, rather than [fiducial markers]({{ site.baseurl }}/programming/sr/vision/markers/).
+
+The `see` method will return a list of visible markers in the arena, but doesn't allow a resolution argument. Each token is as described in the [vision docs]({{ site.baseurl }}/programming/sr/vision/), except:
+
+- `Marker.info.marker_type` will be one of `MarkerType.SILVER`, `MarkerType.GOLD` or `MarkerType.ARENA`. `MarkerType` can be imported with `from sr.robot import MarkerType`
```
PeterJCLaw, line 128: Should we also expose the constants named in the docs?

competition-simulator.md
sedders123: Actually maybe a table would look better with Pin Number and Location being the headers.

competition-simulator.md
antoinepetty: I think it would be best to keep it as `competition-simulator`

competition-simulator.md
antoinepetty: +1 to the table suggestion in the other version of this

competition-simulator.md
RealOrangeOne: I'm not familiar with how that works on windows. I think being explicit like this is still the better option

competition-simulator.md
RealOrangeOne: Have removed, as the editor doesn't show up anymore

competition-simulator.md
RealOrangeOne: I feel like it should be 1 or the other. Either it should be compatible, or it shouldn't.

Personally I like the idea of using enums.

competition-simulator.md
PeterJCLaw: Shebangs don't normally work on Windows, though I don't recall whether Webots implements that limitation. It's possible that it would respect the shebang anyway. I do suggest we check it before removing this though.

competition-simulator.md
PeterJCLaw: I would err on the side of compatibility here. I agree that enums are nice, however I think the closer we can make the code to their existing code the better.

(I'd be tempted to leave the `MarkerType` enum exported anyway, so teams can use it if they want to, just to also add the named constants from the API docs. I realise this violates the Zen of Python, but I think the downsides are small in this case)

competition-simulator.md
antoinepetty: I think Peter's referring to line 52 and 56

competition-simulator.md
RealOrangeOne: https://github.com/srobo/competition-simulator/pull/85

competition-simulator.md
PeterJCLaw: I'm guessing that that @RealOrangeOne's comment here relates to a different thread?

competition-simulator.md
PeterJCLaw: Ah, sorry, I was thinking that we wouldn't necessarily document the enum. I don't think it matters a great deal now it's done though.

competition-simulator/index.md
```diff
@@ -0,0 +1,59 @@
...
+  This is different to the <a href="/docs/programming/simulator/">regular simulator</a>.
+</div>
+
+## Installation
+
+### Prerequisites
+
+This simulation is based in [Webots](https://cyberbotics.com/#download), which will need to be downloaded and installed.
+
+You will also need Python installed. The simulator supports >=3.5. Additional external libraries are not supported.
```
PeterJCLaw, line 22: Maybe link to a suitably page on python.org for convenience?

competition-simulator/index.md
```diff
@@ -0,0 +1,59 @@
...
+
+<div class="info">
+  This is different to the <a href="/docs/programming/simulator/">regular simulator</a>.
+</div>
+
+## Installation
+
+### Prerequisites
+
+This simulation is based in [Webots](https://cyberbotics.com/#download), which will need to be downloaded and installed.
```
PeterJCLaw, line 20: Given the size of the download it might be worth noting that here. I was initially surprised at the size of it and could imagine that others may think they've got the wrong thing.

competition-simulator/index.md
```diff
@@ -0,0 +1,59 @@
...
+
+You may receive a warning about your computer's GPU not being good enough, which can be ignored.
+
+#### Changing your version of Python
+
+If webots is picking up the incorrect version of Python, you'll need to change it. This can be done using `Tools > Preferences > General > Python command`. You'll need to ensure a matching version of Python is installed.
+
+### Updates
+
+Occasionally, we may release an update to the simulation. To update, you will need to delete the `competition-simulator-master` directory, and re-download it using the above link.
```
PeterJCLaw, line 37:
```suggestion
Occasionally, we may release an update to the simulation. To update, you will need to delete the `competition-simulator` directory, and re-download it using the above link.
```

competition-simulator/index.md
```diff
@@ -0,0 +1,59 @@
...
+
+## Overview
+
+Within the Webots IDE, there are a few different panels:
+
+- In the centre of your screen is the 3D simulated view of the arena
+- On the left is a tree hierarchy of all elements in this "world"
+- At the bottom is your console
```
PeterJCLaw, line 45: What is a console? Maybe explain what it's used for?
```suggestion
- At the bottom is the console, where output from your robot code will be displayed
```

PeterJCLaw: Related to this, it occurs to me that there isn't a `log.txt` for teams to review after the fact (currently). That might be worth mentioning here, or alternatively we could introduce a `log-{datetime}.txt` as part of our façade controller.

competition-simulator/index.md
```diff
@@ -0,0 +1,59 @@
...
+- [Camera Controls](https://www.cyberbotics.com/doc/guide/the-3d-window#navigation-in-the-scene)
+- [Graphics settings](https://www.cyberbotics.com/doc/guide/preferences#opengl) (Useful if Webots is running slowly)
+
+## Robot
+
+![]({{ site.baseurl }}/resources/competition-simulator/robot-front.png)
+
+This is the pre-built robot used in the simulator, which supports the SR API. Building robot chassis in Webots is complicated, and we expect teams to use the pre-built robot.
+
+The robot has an array of sensors, a vision system, and a gripper to pick up tokens.
```
PeterJCLaw, line 59: Might be worth explicitly linking from this new page to the programming page.

competition-simulator/index.md
```diff
@@ -0,0 +1,59 @@
...
+### Useful links
+
+- [Camera Controls](https://www.cyberbotics.com/doc/guide/the-3d-window#navigation-in-the-scene)
+- [Graphics settings](https://www.cyberbotics.com/doc/guide/preferences#opengl) (Useful if Webots is running slowly)
+
+## Robot
+
+![]({{ site.baseurl }}/resources/competition-simulator/robot-front.png)
+
+This is the pre-built robot used in the simulator, which supports the SR API. Building robot chassis in Webots is complicated, and we expect teams to use the pre-built robot.
```
antoinepetty, line 57: I feel like this implies it's a possibility to use their own

competition-simulator/programming.md
```diff
@@ -0,0 +1,86 @@
...
+  You will be responsible for backing up and versioning your code, and collaborating with your fellow team members. The <a href="https://studentrobotics.org/ide/">IDE</a> is unsuitable for this simulator.
+</div>
+
+## Programming Interface
+
+Unless otherwise stated, the simulator’s API is the same as the real SR API described in the [programming docs]({{ site.baseurl }}/programming/).
+
+### Motors
+
+Your robot has 2 motor boards attached, each with 2 motors. Board `0` has the left wheel in port `m0`, and the right wheel in `m1`. Board `1` has the gripper lift motor in `m0`, and the finger motors in `m1`.
```
PeterJCLaw, line 38: The rendered visual distinction between the inline 2 and the `0` is small. Suggest using words for numbers which are not code:
```suggestion
Your robot has two motor boards attached, each with two motors. Board `0` has the left wheel in port `m0`, and the right wheel in `m1`. Board `1` has the gripper lift motor in `m0`, and the finger motors in `m1`.
```

competition-simulator/programming.md
```diff
@@ -0,0 +1,86 @@
...
+
+The motor boards do not have part codes, and so need to be indexed using `0` and `1`. The motor boards will always be in this order.
+
+### Ruggeduino
+
+Your robot has 5 microswitches and 6 distance sensors, attached to the digital and analogue pins respectively. These are all attached to a single ruggeduino.
```
PeterJCLaw, line 44: Suggest using words for counts:
```suggestion
Your robot has five microswitches and six distance sensors, attached to the digital and analogue pins respectively. These are all attached to a single ruggeduino.
```

competition-simulator/programming.md
```diff
@@ -0,0 +1,86 @@
...
+
+The microswitches are attached to digital pins 2-6:
+
+| Pin | Location |
+|-----|----------|
+|2|Front|
+|3|Back|
+|4|Between gripper fingers|
+|5|Left gripper finger|
+|6|Right gripper finger|
```
PeterJCLaw, line 58: Would be great for this to visually look like a table in the source too:
```suggestion
|-----|----------|
| 2   | Front    |
| 3   | Back     |
| 4   | Between gripper fingers |
| 5   | Left gripper finger |
| 6   | Right gripper finger |
```

competition-simulator/programming.md
```diff
@@ -0,0 +1,86 @@
...
+Analogous to ultrasound sensors, distance sensors allow you to retrieve the distance between your robot and an object. These are attached to analogue pins 0-5:
+
+| Pin | Location |
+|-----|----------|
+|0|Front Left|
+|1|Front Right|
+|2|Left|
+|3|Right|
+|4|Back Left|
+|5|Back Right|
```
PeterJCLaw, line 73: Would be great for this to visually look like a table in the source too:
```suggestion
|-----|----------|
| 0   | Front Left |
| 1   | Front Right |
| 2   | Left     |
| 3   | Right    |
| 4   | Back Left |
| 5   | Back Right |
```

competition-simulator/index.md
RealOrangeOne: This is actually related to the name of the directory GitHub uses inside the zip file.

Have updated so it's more accurate to what it would be

competition-simulator/programming.md
```diff
@@ -0,0 +1,88 @@
+---
+layout: page
+title: Programming
+---
+
+# Programming
+
+## Developing your code
+
+On first run, the robot will execute an example program for convenience. On first run, this will be copied to the directory `competition-simulator` is stored in:
```
antoinepetty, line 10:
```suggestion
On first run, the robot will execute an example program for convenience. On first run, this will be copied to the directory `competition-simulator-<version>` is stored in:
```

competition-simulator/programming.md
```diff
@@ -0,0 +1,88 @@
...
+
+```
+.
+├── competition-simulator
```
antoinepetty, line 14:
```suggestion
├── competition-simulator-<version>
```

competition-simulator/index.md
PeterJCLaw: https://github.com/srobo/docs/pull/122/commits/e2a57e726bb8fdff650f5ae42b7baf7b110e6125

competition-simulator/index.md
```diff
@@ -0,0 +1,59 @@
...
+- [Camera Controls](https://www.cyberbotics.com/doc/guide/the-3d-window#navigation-in-the-scene)
+- [Graphics settings](https://www.cyberbotics.com/doc/guide/preferences#opengl) (Useful if Webots is running slowly)
+
+## Robot
+
+![]({{ site.baseurl }}/resources/competition-simulator/robot-front.png)
+
+This is the pre-built robot used in the simulator, with an interface [similar](./programming) to the SR API.
+
+The robot has an array of sensors, a vision system, and a gripper to pick up tokens, all of whom can be [programmed](./programming).
```
PeterJCLaw, line 59: The sensors (etc.) aren't people:
```suggestion
The robot has an array of sensors, a vision system, and a gripper to pick up tokens, all of which can be [programmed](./programming).
```

PR #124, opened 2020-06-01
Specify version 3.7

antoinepetty:
Resolves https://github.com/srobo/competition-simulator/issues/89

Webots supports 3.7 on all platforms so we should be explicit in the documentation about this


RealOrangeOne:
That's, strange. I thought @PeterJCLaw had used it on ~3.5

antoinepetty:
> That's, strange. I thought @PeterJCLaw had used it on ~3.5

There's some additional compatibility for specific platforms but 3.7 seems to be the only one which works everywhere https://cyberbotics.com/doc/guide/using-python#introduction

PR #126, opened 2020-06-07
Add documentation on having other robots in the arena

RealOrangeOne:
Document changes from https://github.com/srobo/competition-simulator/pull/99


competition-simulator/index.md
```diff
@@ -67,3 +67,26 @@ Within the Webots IDE, there are a few different panels:
...
+
+There are 3 other robots in the arena, all in their starting zones. By default, these robots will not move.
+
+To control these robots, you will need to give them a controller:
+
+1. Select the robot, and in the menu on the left, edit the controller field and select the `sr_controller` controller:
+
+![]({{ site.baseurl }}/resources/competition-simulator/robot-controller.png)
+
+2. Create the robot code in `zone-x/robot.py`, where `x` is the corner number. Your robot code at `robot.py` will still work.
```
PeterJCLaw, line 81: Hrm, this means that if the team configures a controller for a second robot, but doesn't put the `robot.py` in the right place, they'll get their fallback code used for both.

If they _have_ put another robot file in place, but in the wrong corner, this could be quite confusing.

Should we have the controller output what path it's using for each of the robots?

antoinepetty:
I'm not sure if this is what Peter is saying, but I think we should commit to the `zone-x` directories for all controllers instead of having the single robot.py for robot 0

competition-simulator/index.md
RealOrangeOne: It will use their existing file, yes. That's what I was aiming for it to do, so it's easier to get started with seeing other robots in the arena. Although I guess if they wanted to do that they could just copy the file.

Committing everyone to the `zone-x` spelling would work, but feels weird if they don't want this feature. And then forcing it when they're using multiple robots sounds rather annoying.

PeterJCLaw: I don't feel strongly what the solution here is (and I'm not really advocating for removing the detection of the file location we've already published), but I do think it's important that the behaviour be somewhat "obvious" and easily debuggable.

Here's an alternative idea (though not without its own issues):
- zones 1-3 can only be controlled by their own specific files
- zone 0 can be controlled by `/robot.py` or `zone-0/robot.py`, but it's an error for both to be present
- the _sr_controller_ (rather than the `Robot` class) prints which zones are being set up and which paths its using for each

This keeps compatibility as well as ease of use for development, while enabling teams to make other robots do something if they want them to.

RealOrangeOne: This should now be covered by https://github.com/srobo/competition-simulator/pull/99

PeterJCLaw:
> I'm not sure if this is what Peter is saying, but I think we should commit to the `zone-x` directories for all controllers instead of having the single robot.py for robot 0

That's not quite what I was suggesting, mostly as it would break existing competitors and complicate development in the single-robot case.

I've implemented what I had in mind as https://github.com/srobo/competition-simulator/pull/106. Perhaps let's take the discussion there and then document whatever we come up with?

competition-simulator/index.md
PeterJCLaw: https://github.com/srobo/competition-simulator/pull/106 implements my suggestion above and is merged, shall we go with documenting that?

PeterJCLaw:
```suggestion
2. Create the robot code in `zone-x/robot.py`, where `x` is corner number `1`, `2` or `3`.
   Your code at `robot.py` will be used for zone `0`, though you are encouraged to move your code to `zone-0/robot.py` when running multiple robots.
   Robots which do not have any robot code in their `zone-x` directory will not do anything.
```

PR #138, opened 2020-07-05
The IDE now supports developing for the simulator

PeterJCLaw:
This changed a while back, so remove the warning about it being unsuitable.

### DIFF
--- a/competition-simulator/programming.md
+++ b/competition-simulator/programming.md
@@ -25,10 +25,6 @@ Your code should be developed in `robot.py`.
   Only your controller code will be present in the competition environment.
 </div>
 
-<div class="warning">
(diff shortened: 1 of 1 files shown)


competition-simulator/programming.md
```diff
@@ -25,10 +25,6 @@ Your code should be developed in `robot.py`.
   Only your controller code will be present in the competition environment.
 </div>
 
-<div class="warning">
-  You will be responsible for backing up and versioning your code, and collaborating with your fellow team members. The <a href="https://studentrobotics.org/ide/">IDE</a> is unsuitable for this simulator.
```
RealOrangeOne, line 29: I think we should still keep the message about needing to manually backup and version code, namely when working locally.

PeterJCLaw: I'm not sure what the point is? It feels to me rather like saying "when working locally you need to use your own computer".

PR #144, opened 2020-11-04
Removed references to ManualTimestepRobot from time.md and index.md 

eljueves:


competition-simulator/programming/time.md
```diff
@@ -1,62 +1,23 @@
...
-
-<div class="info">
-  If you find that the simulator freezes then this indicates that your code is
-  busy doing something without giving the simulator a chance to run.
-
-  This usually indicates that you have a loop somewhere which is expecting time
-  to advance on its own and which should be modified to call <code>R.sleep</code>
-  occasionally (even a very small value will allow the simulator to progress).
+  This usually indicates that you have a loop somewhere which is expecting time    
+  to advance on its own and which should be modified to call `sleep`
```
PeterJCLaw, line 19: Unfortunately this will need to stay as HTML tags -- markdown inside other HTML tags (here specifically the `<div>`) isn't parsed.

eljueves: fixed it

competition-simulator/programming/time.md
```diff
@@ -1,62 +1,23 @@
...
-<div class="info">
-  If you find that the simulator freezes then this indicates that your code is
-  busy doing something without giving the simulator a chance to run.
-
-  This usually indicates that you have a loop somewhere which is expecting time
-  to advance on its own and which should be modified to call <code>R.sleep</code>
-  occasionally (even a very small value will allow the simulator to progress).
+  This usually indicates that you have a loop somewhere which is expecting time    
+  to advance on its own and which should be modified to call <code>sleep</code> 
+  occasionally (even a very small value will allow the simulator to progress).
```
PeterJCLaw, line 20: `git` tells me there's a lot of trailing whitespace on these lines (and elsewhere) -- please could you remove that?

I'm guessing that this is an artefact of your editor, rather than something you've done yourself. In general I would recommend configuring your editor to always strip trailing whitespace. Even better if you can configure your editor to support [`EditorConfig`](https://editorconfig.org/) -- many of our repos have suitable `.editorconfig` files which will mean that this sort of thing is handled correctly if your editor is configured suitably to observe that standard.

competition-simulator/programming/time.md
```diff
@@ -1,62 +1,23 @@
 ---
 layout: page
-title: Simulated Time
+title: Robot Time
 ---
 
-# Simulated Time
+# Important methods
```
PeterJCLaw, line 6: Is there a need to change these titles? The topic here hasn't really changed -- it is still the time within the simulator which we're talking about. The fact that the user accesses the time values from the `Robot` object is very much an implementation detail.

competition-simulator/programming/time.md
```diff
@@ -1,62 +1,23 @@
...
-
-<div class="info">
-  If you find that the simulator freezes then this indicates that your code is
-  busy doing something without giving the simulator a chance to run.
-
-  This usually indicates that you have a loop somewhere which is expecting time
-  to advance on its own and which should be modified to call <code>R.sleep</code>
-  occasionally (even a very small value will allow the simulator to progress).
+  This usually indicates that you have a loop somewhere which is expecting time    
+  to advance on its own and which should be modified to call <code>sleep</code>
```
PeterJCLaw, line 19: This line used to say `<code>R.sleep</code>` rather than just `<code>sleep</code>`. The former is more precise and still correct -- what was the motivation for changing it?

competition-simulator/programming/time.md
```diff
@@ -94,72 +55,3 @@ then = R.time()
 now = R.time()
 
 duration = now - then
-```
```
PeterJCLaw, line 97: I think we need to keep this closing code fence?

competition-simulator/programming/time.md
```diff
@@ -1,62 +1,23 @@
...
-# Simulated Time
+# Important methods
 
 In the simulated environment, time advances only at the pace that the simulator
 is run and in cooperation with the controller code (such as the code controlling
-your robot). As a result, using `time.time` to know how long your robot has been
-running for or `time.sleep` to wait for some duration may be unreliable.
+your robot). As a result, using the `time` method to know how long your robot has been
+running for or `sleep` method to wait for some duration is the most reliable way of 
+implementing precise movements in your robot
```
PeterJCLaw, line 12: I'm not sure that this is something that we want to say here. Partly because we don't want to encourage dead-reckoning (i.e: purely time based movements), but also because it may not be true. For example if the robot isn't using its brakes then it is likely that the robot will have travelled a longer than expected amount. Alternatively if competitors use the vision system and/or we offer competitors another locations method (perhaps such as an IMU or wheel position), then using time for precise movements will no longer be the best method.

Instead I suggest that we keep something along the lines of the previous paragraph here, though perhaps strengthen what it says so that it is clear that using the `time` module shouldn't be used.

competition-simulator/programming/time.md
```diff
@@ -1,62 +1,23 @@
...
-
-For new projects, or existing projects that want to be sure of getting precise
-robot behaviour, the recommended approach is to use the `ManualTimestepRobot`
-class.
-
-This approach relies upon your code advancing the simulation explicitly by
-calling its `sleep` method (documented below) in order for the simulation to
-actually run. This should not be an issue for most robot code however as you
-will likely be doing this anyway in order to wait for things to happen.
-
```
PeterJCLaw, line 52: I think we probably do want to keep this paragraph or something like it. While we do allude to this elsewhere, I think the clear idea that this (especially the first sentence) gives -- that the competitors code is directly responsible for the advancement of time and that they can break things if they get it wrong is really important.

(It also sets up the context for the following info box fairly well)

PeterJCLaw (changes requested):
Thanks for putting this together.
I think it needs some changes however before we can merge this, in order to ensure that it conveys all the information that competitors need.

Are you likely to have time to work on this in the next few days? I think this is an important change to make, so if not I'd be happy to pick this up so we can get it in before Kickstart.

eljueves:
@PeterJCLaw I don't think I will have the time, so you can take it on

PeterJCLaw:
No worries, thanks for letting me know @eljueves :)

competition-simulator/programming/time.md
PeterJCLaw: 7ce4d09

competition-simulator/programming/time.md
PeterJCLaw: 1d06526

competition-simulator/programming/time.md
PeterJCLaw: 3059fe8

competition-simulator/programming/time.md
PeterJCLaw: 1d06526

competition-simulator/programming/time.md
PeterJCLaw: b82bc54

competition-simulator/programming/time.md
PeterJCLaw: b126b3c

competition-simulator/programming/time.md
PeterJCLaw: b126b3c

PR #147, opened 2020-11-15
Canonicalise around the new "competition" simulator

PeterJCLaw:
For SR2021 we are running entirely in this simulator and don't really want to have to update the old PyGame based simulator.
Given that the new simulator offers a much better experience anyway, it seems much more reasonable to expect that we're going to use it exclusively in the future.

Dropping the docs and links to the old simulator therefore reduces the work we need to do now and avoids any potential confusion on the part of the competitors.

This therefore also renames references to the simulator to simplify things, but deliberately doesn't change:

- its position in the sidebar
  (its importance this year means it feels to me like it deserves to be both a top-level item and to be at the top of the list)
- the directories we mention that users will see when using it
  (changing this would involve changing the simulator repo itself, which I don't think is needed)


RealOrangeOne (approved):
I suggest we archive the old repository and rename the new one, however that can be done after this PR ships easily enough (GitHub redirects URLs for renamed repositories)

PeterJCLaw:
> I suggest we archive the old repository and rename the new one, however that can be done after this PR ships easily enough (GitHub redirects URLs for renamed repositories)

Yeah, probably. The old repo is [`sr-turtle`](https://github.com/srobo/sr-turtle/) so I don't think we need to rename it?

rgilton:
Hi, thanks for requesting review from the kit team.  The @srobo/support-team is the part of the kit team that is responsible for providing support for the kit, and maintaining its documentation, so please direct your kit documentation/support requests at them :smile:

PR #150, opened 2020-11-15
Add some basic discord docs

RealOrangeOne:
Designed both for reference by competitors and team leaders, as well volunteers.

We don't want to document how to use Discord, just some basics of how ours is setup / should be used.

### DIFF
--- a/team_admin/discord.md
+++ b/team_admin/discord.md
@@ -25,4 +25,4 @@ Along with the common channels, there are some channels visible only to members 
 
 `team-ABC` (where `ABC` is your TLA) is a channel only your team (and blue-shirts) can see. Here you can converse among your team, and get help from blue-shirts without giving away too much of your strategy. These channels should be considered your own, and can be used to discuss anything competition related for your team.
 
-If your school / institution has multiple teams (`ABC2` etc), there will be an additional channel for each of those (`teams-ABC2`). Remember to try and keep conversations specific to the team - Just because the team comes from the same school / institution, doesn't make them any less your rival.
+If your school / institution has multiple teams (`ABC2` etc), there will be an additional channel for each of those (`teams-ABC2`). Remember to try and keep conversations specific to the team - just because the team comes from the same school / institution, doesn't make them any less your rival.
--- a/team_admin/discord.md
+++ b/team_admin/discord.md
@@ -11,7 +11,7 @@ If you do not wish to, or are unable to join Discord, you can still get support 
 
 ## Joining
 
-Invite links will be sent to team leaders which will ensure you're added to the correct school / institution and team. Only team members should be given access. If you do not already have an account, you'll be prompted to create one. You may of course use your existing account, but please make sure the profile picture and username are appropriate.
+Invite links and entry passwords will be sent to team leaders which will ensure you're added to the correct school / institution and team. Only team members should be given access. If you do not already have an account, you'll be prompted to create one. You may of course use your existing account, but please make sure the profile picture and username are appropriate.
 
 Once invited, you'll only have access to a few channels, namely `#welcome-and-rules` and `#welcome-<username>` (with your username in). Before you can join, you must read the rules shown in the `#welcome-and-rules` channel, and enter the password for your team in your personal `#welcome-<username>` channel.
 
--- a/team_admin/discord.md
+++ b/team_admin/discord.md
@@ -5,7 +5,7 @@ title: Discord Server
 
 # Discord Server
 
-To assist with communications, especially in a virtual competition, there's a [Discord](https://discord.com/) communication both for communication with blue-shirts and among your team.
+To assist with communications, especially in a virtual competition, there's a [Discord](https://discord.com/) server for communicating with blue-shirts, your team members and other teams.
 
 If you do not wish to, or are unable to join Discord, you can still get support from <{{ site.emails.teams }}> as normal.
 
--- a/team_admin/discord.md
+++ b/team_admin/discord.md
@@ -11,7 +11,7 @@ If you do not wish to, or are unable to join Discord, you can still get support 
 
 ## Joining
 
-Invite links will be sent to team leaders which will ensure you're added to the correct school / institution and team. Only team members should be given access.
+Invite links will be sent to team leaders which will ensure you're added to the correct school / institution and team. Only team members should be given access. If you do not already have an account, you'll be prompted to create one. You may of course use your existing account, but please make sure the profile picture and username are appropriate.
 
 Once invited, you'll only have access to a few channels, namely `#welcome-and-rules` and `#welcome-<username>` (with your username in). Before you can join, you must read the rules shown in the `#welcome-and-rules` channel, and enter the password for your team in your personal `#welcome-<username>` channel.
 
--- a/team_admin/discord.md
+++ b/team_admin/discord.md
@@ -25,4 +25,6 @@ Along with the common channels, there are some channels visible only to members 
 
 `team-ABC` (where `ABC` is your TLA) is a channel only your team (and blue-shirts) can see. Here you can converse among your team, and get help from blue-shirts without giving away too much of your strategy. These channels should be considered your own, and can be used to discuss anything competition related for your team.
 
+When asking for support, it's best to ask in public rooms unless there's a specific reason not to. This not only allows other teams who may be struggling with the same issue to benefit, but also allows them to contribute to a solution and discuss potential improvements.
+
 If your school / institution has multiple teams (`ABC2` etc), there will be an additional channel for each of those (`teams-ABC2`). Remember to try and keep conversations specific to the team - just because the team comes from the same school / institution, doesn't make them any less your rival.
--- a/team_admin/discord.md
+++ b/team_admin/discord.md
@@ -27,4 +27,4 @@ Along with the common channels, there are some channels visible only to members 
 
 When asking for support, it's best to ask in public rooms unless there's a specific reason not to. This not only allows other teams who may be struggling with the same issue to benefit, but also allows them to contribute to a solution and discuss potential improvements.
 
-If your school / institution has multiple teams (`ABC2` etc), there will be an additional channel for each of those (`teams-ABC2`). Remember to try and keep conversations specific to the team - just because the team comes from the same school / institution, doesn't make them any less your rival.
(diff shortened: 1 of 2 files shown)


team_admin/discord.md
```diff
@@ -0,0 +1,28 @@
...
+
+# Discord Server
+
+To assist with communications, especially in a virtual competition, there's a [Discord](https://discord.com/) communication both for communication with blue-shirts and among your team.
+
+If you do not wish to, or are unable to join Discord, you can still get support from <{{ site.emails.teams }}> as normal.
+
+## Joining
+
+Invite links will be sent to team leaders. Only competitors should be given access.
```
PeterJCLaw, line 14: Should we clarify that invite links are team-specific?

team_admin/discord.md
```diff
@@ -0,0 +1,28 @@
...
+title: Discord Server
+---
+
+# Discord Server
+
+To assist with communications, especially in a virtual competition, there's a [Discord](https://discord.com/) communication both for communication with blue-shirts and among your team.
+
+If you do not wish to, or are unable to join Discord, you can still get support from <{{ site.emails.teams }}> as normal.
+
+## Joining
```
PeterJCLaw, line 12: Do you need to create a Discord account before trying to join our server? Perhaps mention that that's the case if so?

team_admin/discord.md
```diff
@@ -0,0 +1,28 @@
...
+
+The server has a number of channels for specific conversation. Please try to keep channels on-topic, and move channels if the conversation tangents. Each channel has a topic at the top of the screen showing what it's about.
+
+### Team channels
+
+Along with the common channels, there are some channels visible only to members of your team.
+
+`team-ABC` (where `ABC` is your TLA) is a channel only your team (and blue-shirts) can see. Here you can converse among your team, and get help from blue-shirts without giving away too much of your strategy. These channels should be considered your own, and can be used to discuss anything competition related for your team.
+
+If your school / institution has multiple teams (`ABC2` etc), there will be an additional channel for each of those (`teams-ABC2`). Remember to try and keep conversations specific to the team - Just because the team comes from the same school / institution, doesn't make them any less your rival.
```
PeterJCLaw, line 28: Grammar:
```suggestion
If your school / institution has multiple teams (`ABC2` etc), there will be an additional channel for each of those (`teams-ABC2`). Remember to try and keep conversations specific to the team - just because the team comes from the same school / institution, doesn't make them any less your rival.
```
Or break this into two sentences.

team_admin/discord.md
RealOrangeOne: The invite takes the form of a link, which will prompt you to create an account if you don't have one already.

team_admin/discord.md
```diff
@@ -0,0 +1,28 @@
...
+
+## Channels
+
+The server has a number of channels for specific conversation. Please try to keep channels on-topic, and move channels if the conversation tangents. Each channel has a topic at the top of the screen showing what it's about.
+
+### Team channels
+
+Along with the common channels, there are some channels visible only to members of your team.
+
+`team-ABC` (where `ABC` is your TLA) is a channel only your team (and blue-shirts) can see. Here you can converse among your team, and get help from blue-shirts without giving away too much of your strategy. These channels should be considered your own, and can be used to discuss anything competition related for your team.
```
PeterJCLaw, line 26: Are these private to that team? If not, I suggest saying so explicitly as I suspect that teams could easily assume that they are.

Separately, should we provide teams with a private channel? We could presumably set these up so that their team leaders and any volunteer could access it (which I think covers the public/visibility side), but could be useful to the teams.
Feels like this might help teams feel they can discuss their code/strategy with a volunteer more freely, which is the main thing I've always perceived as the reason people don't like using the forums.

team_admin/discord.md
PeterJCLaw: Cool. I think we should say that you'll be prompted to create an account if you don't already have one.

team_admin/discord.md
RealOrangeOne: They are private to teams, yes. The only people who can see those channels are members of that team (competitor or team leader), and blueshirts. 

That's the point of this section (see previous line).

team_admin/discord.md
```diff
@@ -11,7 +11,7 @@ If you do not wish to, or are unable to join Discord, you can still get support
 
 ## Joining
 
-Invite links will be sent to team leaders. Only competitors should be given access.
+Invite links will be sent to team leaders which will ensure you're added to the correct team. Only team members should be given access.
```
PeterJCLaw, line 14: It might be worth mentioning here (in place of or possibly addition to where we say it below) that where there are multiple teams per school there will be a channel per team.

team_admin/discord.md
PeterJCLaw: Turns out: I can't read.

team_admin/discord.md
```diff
@@ -0,0 +1,28 @@
...
+
+## Channels
+
+The server has a number of channels for specific conversation. Please try to keep channels on-topic, and move channels if the conversation tangents. Each channel has a topic at the top of the screen showing what it's about.
+
+### Team channels
+
+Along with the common channels, there are some channels visible only to members of your team.
+
+`team-ABC` (where `ABC` is your TLA) is a channel only your team (and blue-shirts) can see. Here you can converse among your team, and get help from blue-shirts without giving away too much of your strategy. These channels should be considered your own, and can be used to discuss anything competition related for your team.
```
PeterJCLaw, line 26: Counterpoint: how do we ensure that _most_ tech help still happens in public channels? I think it's really important that that remains the case as the visibility of issues that others are having:
- helps others feel it's not just them that are struggling
- means we only need to answer something once, then everyone can see the answer
- avoids some teams getting different levels/amounts of help

RealOrangeOne: I'd say just be explicit. Mention that unless it's sensitive, talk about it in public for the betterment of other teams

PeterJCLaw: :+1: can we put something here about that? (Do let's mention the benefits it brings too)

team_admin/discord.md
```diff
@@ -25,4 +25,6 @@ Along with the common channels, there are some channels visible only to members
 
 `team-ABC` (where `ABC` is your TLA) is a channel only your team (and blue-shirts) can see. Here you can converse among your team, and get help from blue-shirts without giving away too much of your strategy. These channels should be considered your own, and can be used to discuss anything competition related for your team.
 
+When asking for support, it's best to ask in public rooms unless there's a specific reason not to. This not only allows other teams who may be struggling with the same issue to benefit, but also allows them to contribute to a solution and discuss potential improvements.
+
 If your school / institution has multiple teams (`ABC2` etc), there will be an additional channel for each of those (`teams-ABC2`). Remember to try and keep conversations specific to the team - just because the team comes from the same school / institution, doesn't make them any less your rival.
```
PeterJCLaw, line 30: If the channels are private and you can only be in one of them, does the second part of this make sense?

team_admin/discord.md
```diff
@@ -11,7 +11,7 @@ If you do not wish to, or are unable to join Discord, you can still get support
 
 ## Joining
 
-Invite links will be sent to team leaders which will ensure you're added to the correct school / institution and team. Only team members should be given access.
+Invite links will be sent to team leaders which will ensure you're added to the correct school / institution and team. Only team members should be given access. If you do not already have an account, you'll be prompted to create one. You may of course use your existing account, but please make sure the profile picture and username are appropriate.
```
PeterJCLaw, line 14: :+1:

team_admin/discord.md
```diff
@@ -0,0 +1,30 @@
+---
+layout: page
+title: Discord Server
+---
+
+# Discord Server
+
+To assist with communications, especially in a virtual competition, there's a [Discord](https://discord.com/) communication both for communication with blue-shirts and among your team.
```
Adimote, line 8:
```suggestion
To assist with communications, especially in a virtual competition, there's a [Discord](https://discord.com/) chat server both for communication with blue-shirts and among your team.
```

team_admin/discord.md
```diff
@@ -0,0 +1,30 @@
...
+
+If you do not wish to, or are unable to join Discord, you can still get support from <{{ site.emails.teams }}> as normal.
+
+## Joining
+
+Invite links will be sent to team leaders which will ensure you're added to the correct school / institution and team. Only team members should be given access. If you do not already have an account, you'll be prompted to create one. You may of course use your existing account, but please make sure the profile picture and username are appropriate.
```
Adimote, line 14:
```suggestion
Invite links and entry passwords will be sent to team leaders which will ensure you're added to the correct school / institution and team. Only team members should be given access. If you do not already have an account, you'll be prompted to create one. You may of course use your existing account, but please make sure the profile picture and username are appropriate.
```

PeterJCLaw (approved):
Maaaybe worth adding something to the Team Admin index page too?

team_admin/discord.md
Adimote: (mention entry passwords here, because they'll likely be sent at the same time)

team_admin/discord.md
Adimote: This wasn't the plan as I interpreted it? I thought we'd have different roles for the 2 teams, but keep one channel.

Maybe we should just drop this line, we might end up swapping them out based on what teams ask for

PeterJCLaw: If the channels are private then I think it's important that they're private _per-team_. Otherwise it definitely creates an imbalance between teams from single-team-schools and multi-team-schools. I'm not sure what the benefit of school-level-private channels would be?

team_admin/discord.md
Scarzy: Can't suggest a change for some reason:

> To assist with communications, especially in a virtual competition, there's a [Discord](https://discord.com/) chat server both for communication with blue-shirts and among your team.

To assist with communications, especially in a virtual competition, there's a [Discord](https://discord.com/) chat server for communication both with blue-shirts and among your team.

Also, I'd get rid of another 1 of those "communication"'s

team_admin/discord.md
RealOrangeOne: We definitely want different channels per team (and thus need different roles).

in hindsight I agree we don't need the school-level private channel is for multi-team schools.

PeterJCLaw: https://github.com/srobo/docs/pull/150/commits/9142d58583670b7a6987a6ed3c094cccd30bbcf4

PeterJCLaw (approved):

PR #154, opened 2020-11-20
Mechanistic modernisation of the Python code examples

PeterJCLaw:
Pretty much what it says. I've tried to keep this to the code examples.

The main things this does:
- drop using module-formatter (`%`) in favour of `str.format`
- used a context manager for file reading
- wrap some `range` calls in `list`; they're ugly now, but at least match the sample outputs
- change to print calls

See individual commit messages for my reasoning.

PR #169, opened 2020-11-21
More python modernisation

PeterJCLaw:
Couple of small things I missed in #154.

### DIFF
--- a/programming/sr/index.md
+++ b/programming/sr/index.md
@@ -116,7 +116,7 @@ R.wait_start()
 ~~~~~
 
 During the setup phase, the Robot hardware is inaccessible.
-For example,`R.motors` is unavailable since enumerations occurs in the `init` function.
+For example, `R.motors` is unavailable since enumerations occurs in the `init` function.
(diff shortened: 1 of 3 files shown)


sedders123 (approved):

trickeydan:
Again, remit of the support team.

It's worth noting that these PRs are breaking the docs for the physical kit.

PeterJCLaw:
So far all these changes leave the code still functional on 2.7. I wouldn't really want competitors using `input` on Python 2.7, however that seems fairly unlikely to happen on the kit anyway.
The usages of `print` will output different content on Python 2.7, but they will _work_ so I'm not worried by them.

trickeydan:
> So far all these changes leave the code still functional on 2.7. I wouldn't really want competitors using `input` on Python 2.7, however that seems fairly unlikely to happen on the kit anyway.
> The usages of `print` will output different content on Python 2.7, but they will _work_ so I'm not worried by them.

We should have Python 3 on the kit next year anyway, so my point is hopefully moot.

PR #172, opened 2020-11-21
Fix pin location range

RealOrangeOne:
There are only 2 pins, not 5.

As noticed by competitor in https://discord.com/channels/775497131057741836/777582218662510612/779725258080387072


sedders123 (approved):

PeterJCLaw:
Dammit, that's the third PR we've needed to fix this! :shipit:

PR #176, opened 2020-12-23
Distance sensors don't work at extreme angles

WillB97:
Sonar type distance sensor will only detect an object if the pulse contacts the face within a 45 degree cone about the normal.
See https://cyberbotics.com/doc/reference/distancesensor#sonar-sensors for more details.


sedders123 (approved):
This matches both what I've observed in the simulator and the contents of the Webots docs.

RealOrangeOne (approved):

WillB97:
We should also merge #174 at the same time

PR #185, opened 2021-02-04
Specify the version of webots used

RealOrangeOne:
Do not merge until module 3.

We're now changing the version requirement. It'll still work on older ones, but we want to encourage using new ones, as it fixes a number of issues.

### DIFF
--- a/simulator/index.md
+++ b/simulator/index.md
@@ -14,7 +14,7 @@ title: Simulator
 
 ### Prerequisites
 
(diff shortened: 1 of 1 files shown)


WillB97:
Technically the simulator supports R2020b-rev1 and R2021a but competition matches are run on R2021a.
Also errors are only shown in comp mode.

WillB97 (approved):

PR #187, opened 2021-02-06
Add docs page for compass

RealOrangeOne:


PeterJCLaw:
We should also mention within the simulator pages that the robot has a compass.

WillB97 (approved):

Issue #190, opened 2021-02-21
Document simulator noise


PeterJCLaw:
A number of teams have asked about this in Discord. Even though it's not something they can control, we should probably document it.

This might be something worth making general -- so that it's useful to teams in a physical competition too. I think it's ok for that to be secondary though.

PeterJCLaw:
For clarity: I think we should document the presence of noise across all the inputs and outputs that we have it configured for.

PR #191, opened 2021-02-21
Add note about radio noise

RealOrangeOne:
Fixes https://github.com/srobo/docs/issues/190


WillB97 (approved):

PeterJCLaw:
Ah, should have been clearer -- I was assuming that a fix to #190 would include documenting all forms of noise we inject to add realism (though perhaps not audible noise, since we deliberately omit that).

PeterJCLaw:
Perhaps let's get this merged (but leave the issue open) and we can add notes about the other forms of noise later.

PeterJCLaw:
Closing as the noise is now documented. Feel free to re-open if you feel it's important we document the noise in the docs for each sensor/actuator.

PR #195, opened 2021-03-07
Document code editors including VSCode setup

PeterJCLaw:


programming/editors/vscode.md
```diff
@@ -0,0 +1,193 @@
...
+### Configure VSCode
+
+Next, we are going to create a [debug configuration][debug-config] so that
+VSCode knows what you want it to debug.
+
+1. Open the workspace containing your code.
+2. Select the Run view in the sidebar (or press
+   <kbd>Ctrl</kbd> + <kbd>Shift</kbd> + <kbd>D</kbd> /
+   <kbd>⌘</kbd> + <kbd>⇧</kbd> + <kbd>D</kbd> on macOS)
+3. Click "create a launch.json file".
```
Adimote, line 106: If you haven't opened a folder, the blue text is 'open a folder'  ('create a launch.json file' is much harder to spot.

programming/editors/vscode.md
```diff
@@ -0,0 +1,193 @@
...
+4. In the menu which appears, select **Remote Attach**
```
Adimote, line 107: I get a menu asking me which environment to select: 
image

Adimote (commented):
Added some comments on the VSCode process.

programming/editors/vscode.md
PeterJCLaw: Does the first point on this list not cover this?
> 1. Open the workspace containing your code.

programming/editors/vscode.md
PeterJCLaw: Ah, I suspect that if you've got a Python file open (which I probably did) then it'll skip straight to that. Will add an entry about this (and it being optional).
Presumably if you select Python then the rest is the same?

PeterJCLaw: d9240f3b904f66a66f1acd008a09f2034f6f59c1 should fix this.

programming/editors/vscode.md
PeterJCLaw: d9240f3b904f66a66f1acd008a09f2034f6f59c1 should help emphasise this.

programming/editors/vscode.md
```diff
@@ -0,0 +1,203 @@
...
+While VSCode has some built-in support for developing Python, we recommend that
+you additionally install the official [Python extension][ms-python.python]. This
+brings more comprehensive support and enables everything else we will document
+here.
+
+[ms-python.python]: https://marketplace.visualstudio.com/items?itemName=ms-python.python
+
+## Code Completions
+
+In order for VSCode to pick up the `sr.robot` library and offer completions
```
Adimote, line 26: Do you think it's worth linking to the wiki page for code completions here (and syntax highlighting above)? Just in case the user doesn't read the index page.

Adimote (approved):

programming/editors/vscode.md
PeterJCLaw: Hrm, perhaps.

programming/editors/vscode.md
```diff
@@ -0,0 +1,203 @@
+---
+layout: page
+title: VSCode
+---
+
+# Visual Studio Code
+
+Visual Studio Code (known commonly as VSCode) is a free code editor developed by
```
sedders123, line 8: Feel like we should have a link to the VSCode home page on our first mention
```suggestion
[Visual Studio Code](https://code.visualstudio.com/) (known commonly as VSCode) is a free code editor developed by
```

programming/editors/vscode.md
sedders123, line 26: A fresh install of VSCode prompted me to install pylint, if I hit yes the instructions described here still result in a pylint linting error. Should we recommend not to install pylint or include instructions to fix pylint?

sedders123 (commented):
Thanks for putting these docs together Peter!

Followed the guide from scratch in a [Windows Sandbox](https://techcommunity.microsoft.com/t5/windows-kernel-internals/windows-sandbox/ba-p/301849) to get a clean starting point. All instructions made sense to me and could be run successfully.

Just a minor suggestion and a question around interaction with pylint inline below.

programming/editors/vscode.md
PeterJCLaw: 5889517

programming/editors/vscode.md
PeterJCLaw: bef42d7

PeterJCLaw: I agree it's not optimal, however I'm most tempted to ignore this. My reasoning (as discussed in person) is that the popup can also appear whenever the user opens a Python file and is sufficiently out of the way of the flow we're taking them through here that I don't think it's likely to block them from using this. We definitely could walk people through configuring pylint, though I suspect that would fit better in a separate section on linters rather than as part of the editor configs.

PR #201, opened 2021-04-03
Update dependencies

PeterJCLaw:
This picks up a Kramdown security update (in 2.3.1) as well as a variety of other updates. From manual testing the changes in the output are:

- code elements (both blocks and inline) now have the additional CSS class 'language-plaintext' (which has no effect)
- the CSS classes for inline code highlighting have changed, though the appearance is unaffected
- `role` attributes are added to various elements, mostly footnotes

None of these seem to affect the appearance of the docs.


PeterJCLaw:
In case it helps, we're almost certainly already using these versions in production -- here's the versions which GitHub Pages uses: https://pages.github.com/versions/

sedders123 (approved):

PR #211, opened 2021-10-09
Remove out of date shipping information

Tyler-Ward:
The information on this page is quite out of date and we don't have appropriate information to replace it so remove it until better information can be prepared


trickeydan (approved):
We can re-add this page wish the correct information when we have a defined process

PR #212, opened 2021-10-09
Remove Shipping link from sidebar.yml

Tyler-Ward:
The sidebar is manually configured and the link needs to be deleted from it manually


PeterJCLaw (approved):
Change is good, though please could we also remove the link from the main "Team Admin" page?

PR #215, opened 2021-10-17
Add html-proofer as a validation mechanism

PeterJCLaw:
Contributes towards https://github.com/srobo/tasks/issues/701.

Contrary to https://github.com/srobo/website/pull/318, I've used `{{ site.url }}` rather than our literal domain name. This feels like it might be slightly nicer here, though there is some room for confusion between it and `{{ site.baseurl }}`. Would be good to get feedback on this -- if it's preferred I'll update the website repo to do the same.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -31,7 +31,5 @@ task :validate => [:build] do
   # --empty-alt-ignore  # To avoid needing to fix lots upfront, we can migrate towards this later.
   # --allow-hash-href   # Allow empty `#` links to mean "top of page". It's true that these can be errors, however we have far too many to really address this.
   # --url-swap          # Adjust for Jekyll's baseurl. See https://github.com/gjtorikian/html-proofer/issues/618 for more.
-  # --url-ignore        # Allow mailto links without a target email, for our Share links. Works around https://github.com/gjtorikian/html-proofer/issues/552.
-  #                     # Allow old event links which aren't easily updated.
   sh('bundle exec htmlproofer _site --assume-extension --disable-external --empty-alt-ignore --allow-hash-href --url-swap "^/docs/:/"')
 end
--- a/Rakefile
+++ b/Rakefile
@@ -23,3 +23,15 @@ end
 task :build => [:dependencies, :submodules] do
   sh('bundle exec jekyll build --config _config.yml')
 end
+
+task :validate => [:build] do
+  # Explanation of arguments:
+  # --assume-extension  # Tells html-proofer that `.html` files can be accessed without the `.html` part in the url.
+  # --disable-external  # For speed. Ideally we'd check external links too, but ignoring for now.
+  # --empty-alt-ignore  # To avoid needing to fix lots upfront, we can migrate towards this later.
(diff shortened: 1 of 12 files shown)


Rakefile
```diff
@@ -23,3 +23,15 @@ end
...
+task :validate => [:build] do
+  # Explanation of arguments:
+  # --assume-extension  # Tells html-proofer that `.html` files can be accessed without the `.html` part in the url.
+  # --disable-external  # For speed. Ideally we'd check external links too, but ignoring for now.
+  # --empty-alt-ignore  # To avoid needing to fix lots upfront, we can migrate towards this later.
+  # --allow-hash-href   # Allow empty `#` links to mean "top of page". It's true that these can be errors, however we have far too many to really address this.
+  # --url-swap          # Adjust for Jekyll's baseurl. See https://github.com/gjtorikian/html-proofer/issues/618 for more.
+  # --url-ignore        # Allow mailto links without a target email, for our Share links. Works around https://github.com/gjtorikian/html-proofer/issues/552.
+  #                     # Allow old event links which aren't easily updated.
+  sh('bundle exec htmlproofer _site --assume-extension --disable-external --empty-alt-ignore --allow-hash-href --url-swap "^/docs/:/"')
```
raccube, line 36: This is very nice.

Rakefile
```diff
@@ -23,3 +23,15 @@ end
...
+
+task :validate => [:build] do
+  # Explanation of arguments:
+  # --assume-extension  # Tells html-proofer that `.html` files can be accessed without the `.html` part in the url.
+  # --disable-external  # For speed. Ideally we'd check external links too, but ignoring for now.
+  # --empty-alt-ignore  # To avoid needing to fix lots upfront, we can migrate towards this later.
+  # --allow-hash-href   # Allow empty `#` links to mean "top of page". It's true that these can be errors, however we have far too many to really address this.
+  # --url-swap          # Adjust for Jekyll's baseurl. See https://github.com/gjtorikian/html-proofer/issues/618 for more.
+  # --url-ignore        # Allow mailto links without a target email, for our Share links. Works around https://github.com/gjtorikian/html-proofer/issues/552.
+  #                     # Allow old event links which aren't easily updated.
```
RealOrangeOne, line 35: Question: Should there be a documented argument here?

RealOrangeOne (approved):

Rakefile
PeterJCLaw: Hrm, looks like this is a copy/pasta from the website PR. I updated the actually command line but not the docs. cddd33b fixes that.

raccube (approved):

PeterJCLaw:
> Contrary to [srobo/website#318](https://github.com/srobo/website/pull/318), I've used `{{ site.url }}` rather than our literal domain name. This feels like it might be slightly nicer here, though there is some room for confusion between it and `{{ site.baseurl }}`. Would be good to get feedback on this -- if it's preferred I'll update the website repo to do the same.

It seems that `{{ site.url }}` ends up not rendering the full url locally, which seems to rather undermine its purpose :( I therefore don't think we should go with that approach for now.

Issue #235, opened 2021-10-30
Power: complete review of page (it’s r.power_board now too), also add individual output currents


PeterJCLaw:

PeterJCLaw:
7b28f6c603048ebe96c98bfdbe1d255dc47bf5d2

PR #236, opened 2021-10-30
Update power board docs for SR2022

PeterJCLaw:
Fixes https://github.com/srobo/docs/issues/235


programming/sr/power/index.md
```diff
@@ -10,25 +10,9 @@ There are a few things that can be done with the power board, namely current and
 As there is only one power board, it is not accessed like a list like `motors` for example:
```
WillB97, line 10:
```suggestion
As there is only one power board, it is not accessed like a list like `motors` and is instead accessed directly, for example:
```

WillB97 (approved):

PR #267, opened 2021-11-07
Optimise images

raccube:
Saved 274 KB out of 2,3 MB. 18,2 % per file on average (up to 77,9 %)

Done using [ImageOptim](https://imageoptim.com/mac)

<details><summary>ImageOptim Settings</summary>
image

image

image
</details>


images/content/kit/battery_charger_hke4_diagram.svg
```diff
@@ -1,644 +1 @@
...
...ll="none" marker-start="url(#a)" stroke="#000" stroke-width="2.834646"/><path d="m237.40157 797.24396-35.43307-35.43307h-46.06299" fill="none" marker-start="url(#a)" stroke="#000" stroke-width="2.834646"/><text style="font-size:28;letter-spacing:0;word-spacing:0;text-anchor:middle;font-family:Bitstream Vera Sans;writing-mode:lr" x="99.212601" y="733.46448"><tspan font-family="Bitstream Vera Sans" font-size="28" text-anchor="middle" writing-mode="lr" x="99.212601" y="733.46448">Charge</tspan><tspan font-family="Bitstream Vera Sans" font-size="28" text-anchor="middle" writing-mode="lr" x="99.212601" y="773.46448">Status</tspan><tspan font-family="Bitstream Vera Sans" font-size="28" text-anchor="middle" writing-mode="lr" x="99.212601" y="813.46448">LEDs</tspan></text><text style="font-size:28;letter-spacing:0;word-spacing:0;text-anchor:middle;font-family:Bitstream Vera Sans;writing-mode:lr" x="432.28345" y="574.01569"><tspan font-family="Bitstream Vera Sans" font-size="28" text-anchor="middle" writing-mode="lr" x="432.28345" y="574.01569">Power Lead</tspan></text><path d="m265.74803 563.3857h70.86614" fill="none" marker-start="url(#a)" stroke="#000" stroke-width="2.834646"/></g></svg>
```
PeterJCLaw, line 1: The SVGs here are the source files for their diagrams. Suggest that we not "optimise" these source images.
SVGs will be served GZIPped (or better), so they'll end up about the same size as the optimised versions anyway meaning that there's likely no substantial gain here.

images/logo-white.svg
```diff
@@ -1,10 +1 @@
-<?xml version="1.0" encoding="UTF-8"?>
```
PeterJCLaw, line 1: The logo images are likely (i.e: should be) plain copies of images from the style repo. I would encourage that they not be modified here. If you feel compelled to optimise them, I suggest looking at that repo instead.
The value of them being exact copies of the files from the style repo almost certainly outweighs the minor size benefit from this change (especially as they'll be served GZIPPed anyway).

PR #268, opened 2021-11-09
Update the interactive troubleshooter for SR2022

trickeydan:
- Remove the IDE section
- Vision now ignores IDs it doesn't recognise
- Vision is now very fast, except for the first image. The resolution also cannot be adjusted.


resources/troubleshooter/data.json
```diff
@@ -433,7 +429,7 @@
                 },
                 {
                     "answer": "The camera takes a long time to take a photo.",
-                    "message": "Try reducing the resolution in your code, you'll have to find an optimum resolution for your robot to get the best speed and detail."
+                    "message": "The vision system always takes longer for the first photo. Consider taking a photo when your robot first starts."
```
PeterJCLaw, line 432: Is resolution no longer configurable? If it is configurable, then that might still be something we want to mention here.

PeterJCLaw (approved):

resources/troubleshooter/data.json
trickeydan: It is no longer configurable, but performance is sufficiently high that it's not such an issue.

resources/troubleshooter/data.json
```diff
@@ -442,24 +438,24 @@
...
-        "ide": {
-            "question": "Choose the option that best describes your situation.",
+        "vision-markers-dont-get-picked-up-good-webcam": {
+            "question": "Is the marker ID listed in the rules as in use?",
             "answers": [
                 {
-                    "answer": "My code does not check correctly.",
-                    "message": "Click on the 'Error' tab to see detailed explanations of the errors."
+                    "answer": "Yes",
+                    "message": "Make sure the marker is flat, well-lit and there a visible border around the pattern, and if your robot is moving you should check the camera is not shaking.<br />If you still cannot pickup any markers there could be a problem with your code, consider posting on the <a href=\"/forums\" target=\"_blank\">forums</a>."
```
raccube, line 454: Replace forums with Discord.

raccube (changes requested):

PeterJCLaw:
Can we ship this? The forums -> Discord thing doesn't feel like it's a blocker here, especially as the link target does work (and we generally still refer to this as "forums" in the docs).

sedders123 (approved):

raccube (approved):

PR #280, opened 2021-11-12
Override paragraph styling inside message boxes

### LINKED ISSUE #279: Improve styling of info boxes which enable markdown processing
Specially the text inside a:
```md
<div class="info"></div>
```
will be blue, however the text inside:
```md
<div class="info" markdown="1"></div>
```
will be black.

sedders123:
Content within a div that has markdown enabled is wrapped in a p tag which means our custom styles were overwritten.

Fixes #279

### DIFF
--- a/_sass/messages.scss
+++ b/_sass/messages.scss
@@ -14,6 +14,8 @@ $error: #BF433A;
   color: $info;
 
   p {
+    // Prevent our custom styles being overwritten if a div
+    // has markdownenabled="1"
     color: $info;
   }
 }
@@ -25,6 +27,8 @@ $error: #BF433A;
   color: $error;
 
   p {
+    // Prevent our custom styles being overwritten if a div
+    // has markdownenabled="1"
     color: $error;
   }
 }
--- a/_sass/messages.scss
+++ b/_sass/messages.scss
@@ -17,6 +17,7 @@ $error: #BF433A;
     // Prevent our custom styles being overwritten if a div
     // has markdownenabled="1"
     color: $info;
+    margin: unset;
   }
 }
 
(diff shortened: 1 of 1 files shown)


_sass/messages.scss
```diff
@@ -12,11 +12,19 @@ $error: #BF433A;
   background: transparentize($info, 0.8);
   border-color: $info;
   color: $info;
+
+  p {
+    color: $info;
```
RealOrangeOne, line 17: Question: Is this actually a use case where `!important` on the above `color` definition is actually the correct solution

PeterJCLaw: I don't think so, as that would mean that links were blue (rather than orange) too, which we probably don't want.

_sass/messages.scss
```diff
@@ -12,11 +12,19 @@ $error: #BF433A;
   background: transparentize($info, 0.8);
   border-color: $info;
   color: $info;
+
+  p {
```
PeterJCLaw, line 16: Might be worth a comment here (and in `.warning`) about why this is needed?

_sass/messages.scss
sedders123: The more specific selector of `.doc p` means that `!important` does not effect the styling.
image

_sass/messages.scss
sedders123: 👍 Sounds sensible, will add a comment

raccube, line 16: Might also want to remove margin so that it looks the same as the other info boxes.

sedders123: a25d6f4

sedders123: 1cbddd6

raccube (approved):

PR #293, opened 2021-11-18
Updates page

RealOrangeOne:
This adds a dedicated page to list the updates available for a kit, making it simple to download specific versions, and see whether they're on the latest version.

image

It currently doesn't show a changelog. Ideally we'd just link out to the relevant repository's releases, but I couldn't find the ideal one.

Download links are also setup to be tracked with Plausible, so we can see what download rates are.


PeterJCLaw:
@RealOrangeOne should this be in the sidebar too?

_layouts/updates.html
```diff
@@ -0,0 +1,49 @@
+---
```
PeterJCLaw, line 1: Does this page really need to be its own layout? Would an `{% include %}` work?
The latter feels like it's better to me as it would greatly reduce the duplication of the page layout code.

_data/sidebar_tree.yaml
```diff
@@ -75,6 +75,8 @@ tree:
     tree:
       - url: /simulator/programming/
         title: Programming
+  - url: /updates/
+    title: Updates
```
PeterJCLaw, line 79: Would it be better to nest this under the Brain Board page? After all that's what it relates to -- it's not general updates. (Note for example that the Simulator could also have updates, but I'd expect those to be on the Simulator pages).

trickeydan (approved):
We're likely to have a new version later today, so it'd be great to get this shipped

trickeydan (approved):
:tada:

_data/sidebar_tree.yaml
```diff
@@ -17,6 +17,9 @@ tree:
         title: IMAX B6 Charger
     - url: /kit/brain_board
```
PeterJCLaw, line 18: Please can we also update this url so that it matches the actual target url (so that the highlighting of the current page works):
```suggestion
    - url: /kit/brain_board/
```

PR #298, opened 2021-11-20
Update the current Ruggeduino FW_VER to1

trickeydan:
In order to match the version that we flash to boards, this should be `1`.

This needs to be changed as the latest version of the kit software actually verifies that the firmware versions are supported before proceeding.


WillB97 (approved):
From #299 
> The firmware version is no longer used to differentiate stock and extended firmwares so keeping this in step with the stock firmware allows us to keep track of the starting versions of extended firmwares.

trickeydan:
We decided to ship this in the kit team meeting.

PR #299, opened 2021-11-22
Update ruggeduino firmware version to match the default

WillB97:
The firmware version is no longer used to differentiate stock and extended firmwares so keeping this in step with the stock firmware allows us to keep track of the starting versions of extended firmwares.


WillB97:
This is a duplicate of #298

PR #305, opened 2021-12-08
Add a docker container for easier local dev

RealOrangeOne:
This container is only intended for local development!

There's a CI task to ensure the container keeps working, as I think it's the avenue we should recommend for local development.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,6 +4,12 @@ The Student Robotics public documentation.
 
 ## Getting Started
 
+For ease of setup, a Docker container is provided. Simply install Docker and `docker-compose`, and run `docker-compose up`.
+
+Once setup, the site will be accessible on http://localhost:4000/docs/
+
+### Manual
+
 1. [Install Ruby][install-ruby]
 
 2. Install Bundler (1.x) and Rake
--- a/Rakefile
+++ b/Rakefile
@@ -9,7 +9,11 @@ task :deep_clean => [:clean] do
 end
 
 task :dependencies do
-  sh('bundle install --path gems')
+  if ENV["GLOBAL_GEMS"]
+    sh('bundle install')
+  else
+    sh('bundle install --path gems')
+  end
 end
 
 file '_sass/brand/.git' do
@@ -19,7 +23,7 @@ end
 task :submodules => ['_sass/brand/.git']
 
 task :dev => [:dependencies, :submodules] do
-  sh('bundle exec jekyll serve --drafts --config _config.yml')
+  sh('bundle exec jekyll serve --host 0.0.0.0 --drafts --config _config.yml')
 end
 
 task :build => [:dependencies, :submodules] do
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,5 @@
+/gems
+_site
+.sass-cache
+.jekyll-metadata
+.bundle
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -25,3 +25,14 @@ jobs:
 
     - name: Build & Validate
       run: rake validate
+
+  docker:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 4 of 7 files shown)


trickeydan (approved):
Works for me!

Rakefile
```diff
@@ -19,7 +23,7 @@ end
 task :submodules => ['_sass/brand/.git']
 
 task :dev => [:dependencies, :submodules] do
-  sh('bundle exec jekyll serve --drafts --config _config.yml')
+  sh('bundle exec jekyll serve --host 0.0.0.0 --drafts --config _config.yml')
```
PeterJCLaw, line 26: This means that the default local development server will be publicly visible from a developer's machine. I'm not sure what the security implications of that are, but likely not great given that `jekyll serve` is meant for development.

Was this considered when putting this PR together?

README.md
```diff
@@ -4,6 +4,12 @@ The Student Robotics public documentation.
 
 ## Getting Started
 
+For ease of setup, a Docker container is provided. Simply install Docker and `docker-compose`, and run `docker-compose up`.
```
PeterJCLaw, line 7: Suggest that we not put the docker setup first given that installing and setting up docker is a) not detail here at all (unlike the manual docs which include links and guidance for setup), and b) not at all easy (given that it usually involves messing with user group membership and accepting the security trade-offs involved in running docker locally).

README.md
```diff
@@ -4,6 +4,12 @@ The Student Robotics public documentation.
...
+
+Once setup, the site will be accessible on http://localhost:4000/docs/
```
PeterJCLaw, line 9: This section makes no mention of how the user should:
a) update the build and running images when they make changes to the local files they're editing
b) run the validation logic (spell checking, validation of various data files, etc.)

Admittedly the README as it stands is also missing instructions on running `rake validate`, however given the relative simplicity (and tab-completed-ness) of doing so that would be easy to fix. Without an easily accessible development setup, the docker approach would make running that considerably harder.

PeterJCLaw (commented):
Apologies for missing this PR until after it's been reviewed & merged already, however I have some rather substantial concerns about the usability of the docker instructions for a new contributor. They are rather incomplete and the additional layer of indirection they create will make it harder for developers to make any changes that do not immediately work as intended.

Rakefile
RealOrangeOne: It's a development server, but as it's mostly static, security issues should be low. Although I agree yes we might as well add in some protections where it's trivial.

README.md
RealOrangeOne: I think it's a good thing not documenting how to instal docker, as there's nothing project specific in it, and it varies by platform anyway.

As for the ease, it's really not bad nowadays. It's in most package managers, and the desktop version is meant to be simple to use. Things do often get run as `root` however, yes.

README.md
RealOrangeOne: > update the build and running images when they make changes to the local files they're editing

That's because there are no steps. Files are bind-mounted into the running container, so changes on the host are immediately reflected inside the container, and `jekyll` will rebuild them. The process for dependency updates is to rebuild the container, which yes isn't documented.

PeterJCLaw: Ah, I'd missed that it was also bound. I was looking at the `COPY` in the `Dockerfile`. Maybe that should be removed in favour of a comment about the `docker-compose.yml`.

README.md
PeterJCLaw: > I think it's a good thing not documenting how to instal docker, as there's nothing project specific in it, and it varies by platform anyway.
> 
> As for the ease, it's really not bad nowadays. It's in most package managers, and the desktop version is meant to be simple to use. Things do often get run as `root` however, yes.

This may be the case, however there's no links or anything. Therefore for people who are most likely to benefit from not installing Ruby (i.e: people who aren't developers) they're dropped in the deep end.

PR #321, opened 2022-03-12
Add clarification of 'spherical' variables

Minion3665:
It's not currently clear how the diagram maps to the variables; this pull request changes that.


trickeydan (approved):
Thanks for the contribution. Content is correct and approved by me.

We are planning a review of this diagram, but I'd suggest merging this in for now just for clarity.

WillB97:
I agree

PR #322, opened 2022-03-12
Fix link to the ThreeDCoordinate section

PeterJCLaw:
The naming of these objects also seems somewhat confusing -- _lots_ of these are three dimensional coordinates, so the one called `ThreeDCoordinate` should probably be called `CartesianCoordinate`, since that's what it actually is. Obviously that's an API change rather than something we can do now.

### DIFF
--- a/programming/sr/vision/index.md
+++ b/programming/sr/vision/index.md
@@ -157,7 +157,7 @@ spherical
 :   A [`Spherical`](#Spherical) instance describing the position relative to the camera.
 
 cartesian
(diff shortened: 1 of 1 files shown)


WillB97 (approved):
ThreeDCoordinate makes sense since we have Coordinate for the 2D cartesian system.

PeterJCLaw:
> ThreeDCoordinate makes sense since we have Coordinate for the 2D cartesian system.

There's no reason we couldn't have `Polar` 2D coordinates. `ThreeDCoordinate` underspecifies the details and is potentially confusing given that there is also a `Spherical` type.

PR #335, opened 2022-08-11
Remove mentions of None from servo docs

tpoomlmly:
The V4 servo board doesn't support unpowered servos

### DIFF
--- a/programming/sr/cheat_sheet.md
+++ b/programming/sr/cheat_sheet.md
@@ -138,13 +138,6 @@ R.servo_board.servos[1].position = 1
 
 You can also set the position to `0`, which is the approximate centre.
 
-This is different to setting the position to `None`, which will unpower the servo.
-
-~~~~~ python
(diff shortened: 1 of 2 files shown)


trickeydan:
The current firmware on the v4 servo boards does not support this. (Surprised this was not raised by a team!)

@WillB97 Please could you confirm that the new firmware will support this?

I suggest keeping this PR open until we have confirmation that the new firmware is being deployed.

WillB97:
See `SERVO:<n>:DISABLE` of the [interface specification](https://docs.google.com/spreadsheets/u/0/d/1rYcP13dElJRiGtB3a_GBeYGxOcN54JenzaGxBPLc6js/notify/show?s=69c85ec7230ed9dc&rev=573&gid=0).

trickeydan (approved):

trickeydan:
This functionality isn't on the roadmap until SR2024. Thanks Tommy!

PR #338, opened 2022-09-11
Split the brain board update docs into a separate page.

trickeydan:
We will be adding extra content to both the brain board and this new page soon, so let's split them up before the pages get too long.

### DIFF
--- a/kit/brain_board/updating-the-brain.md
+++ b/kit/brain_board/updating-the-brain.md
@@ -1,0 +1,36 @@
+---
+layout: page
+title: Updating The Brain
+---
+
+{% assign latest_version = site.data.kit_versions | first %}
+{% assign latest_version = latest_version.version | default: "2022.1.0" %}
+
+# Updating The Brain
+
+### Flashing SD card
+
+In order to get access to the SD card, you will need to open up the case for the brain board. Pull the latch at the end of the case in order to hinge it open.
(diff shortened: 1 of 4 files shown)


kit/brain_board/updating-the-brain.md
```diff
@@ -0,0 +1,36 @@
+---
```
RealOrangeOne, line 1: Suggestion: The updates list page is pretty small - perhaps these update instructions should go there instead? That way there's a single page for "what are the updates" and "how do I install them".

trickeydan: This page will get much longer when we add info about the file based update system.

kit/brain_board/updating-the-brain.md
```diff
@@ -0,0 +1,36 @@
...
+layout: page
+title: Updating The Brain
+---
+
+{% assign latest_version = site.data.kit_versions | first %}
+{% assign latest_version = latest_version.version | default: "2022.1.0" %}
+
+# Updating The Brain
+
+### Flashing SD card
```
sedders123, line 11: Does the heading structure still make sense now its on its own page?

I would have thought Flashing SD Card would be h2, Etcher h3 and Flashing h4.

trickeydan:
We're dropping rauc this year.

Issue #342, opened 2022-09-15
API quick reference


8BitJosh:
I think the api quick reference should have the following line of code added

from sr.robot3 import *

PeterJCLaw:
Could you clarify which page you mean? (I don't think there's one called "API quick reference") Perhaps link to the page?

I initially thought you might mean https://studentrobotics.org/docs/programming/sr/, yet that does already mention `from sr.robot3 import *` within the first few paragraphs.

WillB97:
https://studentrobotics.org/docs/programming/sr/cheat_sheet has a title of "SR API QUICK REFERENCE".

PeterJCLaw:
Apologies I completely failed to see that.

PR #343, opened 2022-09-16
Bootstrap cheat sheet users with API import

PeterJCLaw:
Fixes https://github.com/srobo/docs/issues/342

### DIFF
--- a/programming/sr/cheat_sheet.md
+++ b/programming/sr/cheat_sheet.md
@@ -9,6 +9,14 @@ This page contains a quick guide to the `sr.robot3` API.
 
 For more information, make sure you check the rest of the documentation.
 
+## Import the API
+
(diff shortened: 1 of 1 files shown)


programming/sr/cheat_sheet.md
```diff
@@ -9,6 +9,14 @@ This page contains a quick guide to the `sr.robot3` API.
 
 For more information, make sure you check the rest of the documentation.
 
+## Import the API
```
PeterJCLaw, line 12: Very open to better names for this -- obviously we know what this means, but I'm not sure that someone new to programming is going to thing "I need to import the API". Ideally feels like this would use terms that a newbie would think to look for.

RealOrangeOne (approved):

Issue #344, opened 2022-09-16
Update 'Getting code on the robot' section


raccube:
In the 2023 Kit Software, [we no longer require robot code to be zipped](https://github.com/srobo/astoria/pull/157).

PR #345, opened 2022-10-02
Update code instructions for removal of robot.zips

PeterJCLaw:
This year we've moved to just running the code directly on the USB stick, without a robot.zip wrapper.

I've eyeballed the menus on macOS, I'm relying on memory for what they're called on Windows.

Fixes https://github.com/srobo/docs/issues/344


antoinepetty (approved):
Feels quite basic but it is correct for Windows

trickeydan (approved):
At a later date we can document how to change the entrypoint, although this is good for now.

raccube:
A couple things:
- The context menus on Windows 11 are a bit different now. I'll get you a screenshot later.
- I'm going to merge in the browser interface soon so that note on restarting code will need to be updated.

PeterJCLaw:
> * The context menus on Windows 11 are a bit different now. I'll get you a screenshot later.

Ah, ok. Thanks.

> * I'm going to merge in the browser interface soon so that note on restarting code will need to be updated.

Oh, is auto-restarting no longer going to be a thing?

PR #348, opened 2022-10-02
Standardise around 5 tilde characters (~)

PeterJCLaw:
While this is a little unusual for markdown these days, having a consistent style is better than a mixture as it enables easier searching for code blocks.


antoinepetty:
Why are we using tilde instead of backticks? I haven't seen this before

PeterJCLaw:
> Why are we using tilde instead of backticks? I haven't seen this before

I think this is a legacy from the version of markdown which was supported in the previous PHP iteration of our site. We could change it, though my goal here was to standardise on the most common form (which is 5 tildes by a large margin).

antoinepetty (approved):

PR #349, opened 2022-10-02
Move the SR2022 rules to the games archive

PeterJCLaw:
We don't want to publish the SR2023 rules yet, so this leaves the rules page with a notice about when they will be published.

Relates to https://github.com/srobo/tasks/issues/952 (see also https://github.com/srobo/recurring-tasks/pull/66).


rules/archive.md
```diff
@@ -13,6 +13,27 @@ current competition year are available from the [main rules page][rules-page].
...
+<img src="{{ site.baseurl }}/images/content/rules/arena-2022.png" alt="A diagram of the arena for the SR2022 game" class="left" />
+
+
+The SR2022 game, _This Way Up_, challenged teams to collect the most tin cans
+from the arena and ensure that they were up the right way for maximum points.
+When in a team's scoring zone, correctly oriented cans were worth 3 points,
+upside down cans were worth one point, and cans lying on their side were worth
+nothing.
+
+However, not all cans are created equal: Tin cans on the arena floor started
```
sedders123, line 28:
```suggestion
However, not all cans were created equal: Tin cans on the arena floor started
```

sedders123 (approved):

PR #351, opened 2022-10-02
Add validation of the interactive troubleshooter and fix urls issue

PeterJCLaw:
This adds validation of the internal question links and of the urls that the troubleshooter contains which point to pages within the docs.
We cannot reasonably validate links outside the docs, so we don't attempt to.

Previously the links had an extra `docs/` in them, which meant that none of them worked either locally or in production.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -1,3 +1,4 @@
+require 'set'
 require 'yaml'
 
 task :clean do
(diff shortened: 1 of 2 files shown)


Rakefile
```diff
@@ -52,6 +53,64 @@ task :validate_kit_versions do
...
 
+task :validate_interactive_troubleshooter_questions => [:build] do
+  # Check that each of the next_question ids actually exists and that each
+  # question is actually used.
+
+  ROOT_QUESTION_ID = ''
+  data = YAML.load_file('resources/troubleshooter/data.json')
+  questions = data['questions']
+
+  targets = Set[]
```
raccube, line 64:
```suggestion
  targets = Set.new
```

The syntax really threw me off

Rakefile
```diff
@@ -52,6 +53,64 @@ task :validate_kit_versions do
...
+    else
+      return if File.file?("_site#{url}.html")
+      raise "Invalid target url '#{url}' in interactive troubleshoter (did you mean '#{url}/'?)\n\n"
+    end
+  end
+
+  text = IO.read('resources/troubleshooter/data.json')
+  text.scan(/href=\\"([^"]+)\\"/) do |match|
+    url = match[0]
+    if url.include? 'ROOT_URL' then
```
raccube, line 106:
```suggestion
    if url.include? 'ROOT_URL'
```

raccube (changes requested):
Tiny un-Ruby-ish things

raccube (approved):

Issue #334, opened 2022-08-02
Update Brain Board Info


trickeydan:
- [x] https://github.com/srobo/docs/issues/340 
- [x] Update kit assembly instructions (#369)
- [x] #380
- [x] Update Updates Page (#337)
- [x] Update Brain Board Page (#355)
- [x] Add LED explanations (#355)
- [x] ~#338~
- [x] Remove all references to ODROID (#355)
- [ ] Update Power board docs to say That L2 is for powering the brain. (#375

Issue #340, opened 2022-09-14
Update kit assembly diagram with new Brain Board


trickeydan:
We need to update the kit assembly diagram with the new brain board

trickeydan:
@Adimote has kindly offered to update his diagram with this. Thanks Andy!

PeterJCLaw:
@Adimote do you know when you're likely to have time to put together the updated diagram? If not, I might have a go later this week.

PR #355, opened 2022-10-02
Update ODROID -> Raspberry Pi

trickeydan:
Relates to https://github.com/srobo/docs/issues/334


kit/brain_board/index.md
```diff
@@ -13,21 +13,53 @@ The below will automatically calculate the latest version number, to be used whe
...
+| 5V Power         | The board is powered    
+| 12V Power        | The board is powered    
+| Reverse Polarity | The 12V power is reversed
+| Boot Progress    | Progress Bar for Brain Boot Progress 
+| Code             | A USB containing code is plugged in
+| Comp             | The Robot is in Competition Mode
+| WiFi             | The Robot WiFi hotspot is running
+| ♥ (Heartbeat)    | Blinks when the Brain is running
+| Start            | The Robot is waiting to start
+| OK               | Shows the code status, see below table
```
PeterJCLaw, line 38: Maybe include the state these would be in after boot, as we do for other boards?

.spelling
```diff
@@ -8,7 +8,6 @@ Ebuyer
 iMAX
 IntelliJ
 Nikitin
-ODROID
```
PeterJCLaw, line 11: I suspect there may be some references to the board in the interactive troubleshooter. Not sure if those get checked by the normal spellchecking stuff.

kit/brain_board/index.md
trickeydan: I think we'd need to define the cutoff a bit more carefully. For many of these LEDs, it depends what's plugged into the Robot. For example, if a code USB was plugged in, the Code LED would be turned on before the Boot Progress bar finishes.

.spelling
PeterJCLaw: Scratch that, grerpping doesn't find anything.

trickeydan: I grepped the whole repo for ODROID, almost everything refers to it (correctly) as the Brain.

PeterJCLaw:
I'm guessing the lack of details in the SVG on GitHub is just GitHub's bad rendering? (The PNG looks good)

trickeydan:
> I'm guessing the lack of details in the SVG on GitHub is just GitHub's bad rendering? (The PNG looks good)

Yep, PNG was rendered from the SVG. SVG is included so that we can edit the PNG easily in the future

kit/brain_board/index.md
PeterJCLaw: I wonder if there's something we can do here though -- for example knowing whether or not the WiFi just hasn't come up _yet_ and you should wait longer for it vs something meaning that it's not going to come up is potentially useful to know.

PeterJCLaw:
Maybe worth sending the PNG through [TinyPNG](https://tinypng.com/)? (and then maybe force-push so the big one isn't in the history)

PeterJCLaw (approved):

kit/brain_board/index.md
trickeydan: Theoretically, the final boot progress LED only turns on when all of the Astoria daemons are alive and working. I *think* in the case of `astwifid`, it only reports as running once it's determined whether to turn on the hotspot. (i.e, the WiFi LED light is on anyway)

Side note: There are tentative plans to get astwifid to connect to the hostapd control socket to check that a) hostapd is actually running; and b) allow us to only set the LED to solid once a client device has connected

trickeydan:
> Maybe worth sending the PNG through [TinyPNG](https://tinypng.com/)? (and then maybe force-push so the big one isn't in the history)

Done

WillB97:
Where were we going to cover that the board is powereed over 12V (supplied from L2 on the power board)?

kit/brain_board/index.md
```diff
@@ -13,21 +13,53 @@ The below will automatically calculate the latest version number, to be used whe
...
+The Brain Board provided with your kit is what runs the code you write and controls the other peripheral boards. It consists of a [Raspberry Pi 4B](https://www.raspberrypi.com/products/raspberry-pi-4-model-b/) and a Student Robotics KCH HAT.
 
-### Flashing SD card
+The LEDs on the HAT display the current status of the robot and can be used to help debug your robot. There are also 3 RGB (Red, Green and Blue) LEDs that you can control from your code.
 
-In order to get access to the SD card, you will need to open up the case for the brain board. Pull the latch at the end of the case in order to hinge it open.
-The SD card is located on the underside of the board. Carefully open the Brain Board's plastic case, grab the SD card with your fingers and simply pull it out of the slot.
+## Board Diagram
+
+![brain board assembly diagram]({{ site.baseurl }}/images/content/kit/brain-diagram.png "The Brain Board Assembly")
```
RealOrangeOne, line 23: Suggestion: We should just embed the SVG version here, instead of adding an extra PNG to the repo. Client coverage is very good for this: https://caniuse.com/svg-img

trickeydan: The SVG doesn't render properly in Chrome or Firefox. I've already tried this.

trickeydan:
> Where were we going to cover that the board is powereed over 12V (supplied from L2 on the power board)?

I would guess wherever we document plugging the 5V power into the ODROID currently? This page didn't cover that before, I was just matching the previous scope of the page.

PeterJCLaw:
> > Where were we going to cover that the board is powereed over 12V (supplied from L2 on the power board)?
> 
> I would guess wherever we document plugging the 5V power into the ODROID currently? This page didn't cover that before, I was just matching the previous scope of the page.

The [kit assembly](https://studentrobotics.org/docs/kit/assembly) page mentions the connections which are needed. For the Brain Board it mentions a provided cable, is that still the case?

The kit assembly page also includes what looks like an ODROID case in its assembly diagram; might be worth seeing if we can update that too.

PeterJCLaw:
> The kit assembly page also includes what looks like an ODROID case in its assembly diagram; might be worth seeing if we can update that too.

Ah, looks like that's handled by https://github.com/srobo/docs/issues/340.

PeterJCLaw:
As there are other issues for updates to the other pages I'm going to merge this.

Issue #327, opened 2022-03-26
Update footer to match website


sedders123:
See https://github.com/srobo/website/pull/360

PR #356, opened 2022-10-02
Sync the site footer with the main website

PeterJCLaw:
This essentially applies https://github.com/srobo/website/pull/360 and any other recent changes.

Fixes https://github.com/srobo/docs/issues/327

### DIFF
--- a/_includes/footer.html
+++ b/_includes/footer.html
@@ -1,13 +1,9 @@
 <footer class="site-footer">
   <div class="row fixed-width">
+    <h4>Student Robotics</h4>
+  </div>
+  <div class="row fixed-width">
     <div class="column d-4-12 m-2-12">
-      <h4>For Competitors</h4>
-      <ul class="sitemap">
-        <li><a href="{{ site.url }}/docs/">Documentation</a></li>
-      </ul>
-    </div>
-    <div class="column d-4-12 m-2-12">
-      <h4>Student Robotics</h4>
       <ul class="sitemap">
         <li><a href="{{ site.url }}/compete/">Compete</a></li>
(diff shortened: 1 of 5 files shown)


_includes/footer.html
```diff
@@ -48,8 +49,11 @@ <h4>&nbsp;</h4>
     <div class="column d-6-12 text-right">
       <ul class="social-media-links">
         {% for account in site.social %}
-        <li><a href="{{ account.link }}"><i class="fa fa-{{ account.icon }}" aria-hidden="true"
-              title="{{ account.title }}"></i></a></li>
+        <li>
+          <a aria-label="{{ account.title }}" href="{{ account.link }}">
+            <i class="fa-brands fa-{{ account.icon }}" aria-hidden="true" title="{{ account.title }}"></i>
```
sedders123, line 54: The `fa-brands` change here was to allow us to upgrade to FontAwesome 6 so we could add TikTok to the social links. Do we change back to `fa` here or also upgrade to FontAwesome 6 on the docs site + add TikTok to socials as part of this PR?

sedders123: For reference https://github.com/srobo/website/pull/434/files

PeterJCLaw: Ah, right. I'll change this back to `fa` and we can handle that separately unless you think it would be easy to roll in?

sedders123: Think the upgrade might be worth its own PR. Changes may required on other pages/parts of the site

PeterJCLaw: #357 handles the font-awesome update.

sedders123 (approved):

Issue #358, opened 2022-10-02
API for controlling KCH LEDs


PeterJCLaw:

WillB97:
The API has since changed

PR #359, opened 2022-10-02
Document how to use the LEDs on the KCH HAT

PeterJCLaw:
I've opted to name this document after the LEDs it controls, rather than the board where those LEDs are hosted, to follow the naming pattern of the "motors", "servos" (etc.) docs which do the same for the other boards. This should be easy to change if naming this after the host board would be preferred.

This is based on the current state of the API integration in https://github.com/srobo/sr-robot3/pull/60 and with reference to [`kch.py` in J5](https://github.com/j5api/j5/blob/main/j5/boards/sr/kch.py), so may not be quite how this is intended to function. Hopefully there's a useful starting point here though.

Fixes https://github.com/srobo/docs/issues/358

### DIFF
--- a/programming/sr/leds/index.md
+++ b/programming/sr/leds/index.md
@@ -9,7 +9,7 @@ LEDs
 The KCH HAT provides diagnostic information about the state of your robot
 through a collection of preconfigured and user-controllable LEDs. It is part of
 the [Brain Board](/docs/kit/brain_board) assembly, however control of its LEDs
-is provide through a distinct API.
+is provided through a distinct API.
 
 The LEDs on the KCH can be accessed via the `kch` object:
 
--- a/programming/sr/leds/index.md
+++ b/programming/sr/leds/index.md
@@ -47,3 +47,15 @@ R.kch.c.rgb = (1, 1, 0)
 # Set LED B to a light blue colour
 R.kch.b.rgb = (0, 1, 1)
 ~~~~~
+
+You can also access all the LEDs together:
+
+~~~~~ python
+# Turn on the blue channel of all the LEDs
+for led in R.kch.leds.values():
+    led.blue = 1
+
+# Set all LEDs to a purple colour
+for led in R.kch.leds.values():
+    led.rgb = (1, 0, 1)
+~~~~~
--- a/programming/sr/leds/index.md
+++ b/programming/sr/leds/index.md
@@ -45,5 +45,5 @@ Alternatively you can set the red, green and blue channels for a given LED all t
 R.kch.c.rgb = (1, 1, 0)
 
 # Set LED B to a light blue colour
-R.kch.b.green = (0, 1, 1)
+R.kch.b.rgb = (0, 1, 1)
 ~~~~~
--- a/programming/sr/leds/index.md
+++ b/programming/sr/leds/index.md
@@ -1,0 +1,49 @@
+---
+layout: page
+title: LEDs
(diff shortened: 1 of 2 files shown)


programming/sr/leds/index.md
```diff
@@ -0,0 +1,49 @@
...
+title: LEDs
+---
+
+LEDs
+====
+
+The KCH HAT provides diagnostic information about the state of your robot
+through a collection of preconfigured and user-controllable LEDs. It is part of
+the [Brain Board](/docs/kit/brain_board) assembly, however control of its LEDs
+is provide through a distinct API.
```
trickeydan, line 12:
```suggestion
is provided through a distinct API.
```

programming/sr/leds/index.md
```diff
@@ -0,0 +1,49 @@
...
+[LEDs](#leds) {#leds}
+---------------------
+
+The KCH HAT has three RGB LEDs: A, B and C. The LEDs default to off, however
+once set they will hold their value even if your code exits. This is potentially
+useful to understand the current state of your code as it runs or the final
+state of the code afterwards.
+
+Each channel of each LED acts independently, so you can either set them separately:
+
```
trickeydan, line 29: Might be worth adding here that although the value assigned is a float between `0` and `1`, we currently don't do PWM so it jumps.

trickeydan (commented):
We might document the `R.kch.leds` attribute too, although I think it'd be better to amend these docs after that after / if we expose the KCHLED Enum. The dict is useful in case competitors want to iterate over the LEDs

trickeydan:
Thanks for writing this Peter! A great start and definitely in a better state than the current implementation of this!

PeterJCLaw:
> We might document the `R.kch.leds` attribute too, although I think it'd be better to amend these docs after that after / if we expose the KCHLED Enum. The dict is useful in case competitors want to iterate over the LEDs

Ah, good point; I'd actually missed that. I'll add that for now as I think you're right it would be useful. Removing it should be easy enough if we decide against it unless you feel we should not add it now?

programming/sr/leds/index.md
PeterJCLaw: I wasn't planning to document that it was a float; just leave the examples as `int`s.
Alternatively, is there a reason that the API doesn't just take `bool`s?

programming/sr/leds/index.md
```diff
@@ -0,0 +1,49 @@
...
+~~~~~
+
+Alternatively you can set the red, green and blue channels for a given LED all together:
+
+~~~~~ python
+# Set LED C to a yellow colour
+R.kch.c.rgb = (1, 1, 0)
+
+# Set LED B to a light blue colour
+R.kch.b.green = (0, 1, 1)
```
sedders123, line 48: Should this be?
```suggestion
R.kch.b.rgb = (0, 1, 1)
```

PeterJCLaw:
> definitely in a better state than the current implementation of this!

Oh, sorry, hope I've not duplicated things here?

programming/sr/leds/index.md
PeterJCLaw: Yes :facepalm:

trickeydan:
> Oh, sorry, hope I've not duplicated things here?

I meant implementation as in the code in `sr.robot3` 😛

WillB97 (approved):

Issue #101, opened 2019-03-31
Document available range of the servo board


PeterJCLaw:
In response to a team having issues with a servo that our kit was unable to use to its full range @richardbarlow posted in the forums:
> ... The timer used to generate the pulses is indeed running at 200kHz, giving a resolution of 5µs. However, it's a 16-bit timer, which means that the pulse width can actually be set up to 65535x5µs (not that such a large pulse width is particularly useful in this case).
>
> The firmware takes the value specified via the robot python library (in the range -100 to 100, inclusive) and adds that onto a 'centre offset' of 300. This gives a range of 200 to 400 timer ticks, or 1ms to 2ms with a centre position of 1.5ms.
>
> Unfortunately there are no standards for servo control signals. It's a world of 'industry standards' (which means it's basically random as to what a given servo expects and a given servo driving source, such as an RC receiver, will output). The range of 1ms to 2ms was chosen based upon the most common pulse width range noted in various placed on the internet at the time of writing the firmware. However, SR is in a somewhat unique position where there is a vast army of people using all sorts of servos with our servo board, which actually provides us with more data to work with.
>
> This issue you are facing how has been encountered one or two times in the past few years and I think it's probably time to try to come up with a solution. Unfortunately this will require a firmware updated to be shipped to to the servo board. The infrastructure is in place to perform this update remotely (as in it will be applied automatically once it has shipped and you download a new robot.zip from the IDE). However, we first need to make the necessary changes to the firmware and test it before shipping the update. We also need to consider backwards compatibility with everyone else's code (as in someone who is currently setting their servo to position '100' would expect this to still be the same after an update).

While we'd ideally like to fix the problem, at least documenting it for now would mean future teams knew what servos to get.

PR #362, opened 2022-10-02
Document the Servo Board control scheme and specifications

PeterJCLaw:
Fixes https://github.com/srobo/docs/issues/101 (from which it also draws its information).

Reviewer: please consider whether the terminology here is useful/correct; I've essentially made it up as I don't know what to call this.

### DIFF
--- a/kit/servo_board.md
+++ b/kit/servo_board.md
@@ -52,6 +52,8 @@ Specification
 | Nominal input voltage                    | 11.1V ± 15% |
 | Output voltage                           | 5.5V      |
 | Maximum total output current [^1]        | 10A       |
+| Servo PWN range                          | 1ms — 2ms |
+| Servo PWN precision                      | 5µs       |
 
 [^1]: If the auxiliary input is connected, outputs 8-11 have an independent maximum current.
 
--- a/kit/servo_board.md
+++ b/kit/servo_board.md
@@ -43,6 +43,30 @@ Case Dimensions
 
 The case measures 68x68x21mm. Don't forget that the cables will stick out.
 
+Servo Control
+-------------
+
+Control of servos is often referred to as using
+<a href="https://en.wikipedia.org/wiki/Pulse-width_modulation">pulse-width modulation (<abbr title="Pulse-width_modulation">PWM</abbr>)</a>
+and while that is a useful way to convey the general idea it is unfortunately
+not a technically precise description of either how the control protocol
+operates nor how the Servo Board implements the control.
+
+Instead RC servos typically to react to the duration of the pulse sent to them
+and essentially ignore the gaps between the pulses. However this is not
+universally the case and some servos do care about the time between pulses as
+well.
+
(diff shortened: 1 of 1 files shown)


WillB97 (commented):
Seems good, just needs the acronyms corrected

richardbarlow:
I feel that using the term PWM for hobby servo control signals is a bit misleading. PWM is generally used to indicate a fixed frequency signal with varying duty cycle. The hobby servo control signal consists of controlled on duration and (mostly) uncontrolled off duration. Therefore it's not a fixed frequency or duty cycle for a given position. It might be more clear to refer to these values as 'Servo control pulse range' and 'Servo control pulse resolution'.

In addition, it would be good to indicate the other timing properties of the signal such as min/max repetition rate. This range is a function of all servo output signals, but it shouldn't be too hard to determine min/max from code inspection.

WillB97:
I agree, something around servo pulse range is a better name.

richardbarlow:
I've just read the code (I didn't leave myself many hints!). I have determined that the minimum overall repetition period is 16.825ms (59.435Hz) and the maximum is 17.825ms (56.101Hz). These are the worst case values for the complete round of 12 pulses, regardless of what combination of servo positions are being output.

However, since the pulses are not always output in the same order each round (see below), the actual observed worse case for a single output needs to take into account a servo going from being the shortest to the longest pulse (which could happen even if that servos position isn't itself changed - all of the other servo positions might have been changed).

If it went from being longest (last output) to shortest (first output) and assuming that it's set to the shortest pulse duration then its repetition rate will be 16.075ms. If it went from shortest to longest and assuming it's set to the longest pulse duration then its repetition rate will be 18.575ms.

Based on all this you could state that the worst case repetition rate values that a single servo will experience are:

Minimum: 16.075ms (62.208Hz)
Maximum: 18.575ms (53.836Hz)

The actual value is very much dependent on the pulse durations of all of the other outputs, but this should not matter at all to the servo. Hence why it's not PWM in the usual sense.

--- 

The reason for the above complexity is due to how the servo board outputs its pulses: The firmware first sorts its output pulses from shortest to longest and then outputs them in a staggered manner. It then sets each output high with a small fixed delay between each output (75us) and finally turns them off one at a time after the necessary delay. Since they were sorted shortest to longest, there is always at least 75us between any required state change. This prevents any interaction between servo outputs - there should be no jitter. The whole process repeats (including the sorting) for each round of pulses.

PeterJCLaw:
Thanks for those details @richardbarlow :)

I've added a new section to the docs to include some of this as background (which I suspect will be useful for everyone) and updated the specifications table with the naming and details as suggested.

Would be great to check if what I've put is both accurate as well as understandable by a lay reader.

richardbarlow (approved):
LGTM. Perhaps if I get a chance at some point I'll make a timing diagram explaining all of this more clearly and it can live next to the firmware

PR #369, opened 2022-10-14
Update kit assembly instructions for the new Brain Board

PeterJCLaw:
This updates the assembly page with a new diagram and instructions since the change from ODROID to Pi+KCH Brain board.

Relates to https://github.com/srobo/docs/issues/334

PR #375, opened 2022-10-18
Mention that the brain board must be powered using L2

RealOrangeOne:


kit/brain_board/index.md
```diff
@@ -24,12 +24,12 @@ The LEDs on the HAT display the current status of the robot and can be used to h
```
RealOrangeOne, line 24: We currently don't mention on the brain board page that it needs to be L2, nor do we mention how to power it. Do we think this is worth doing, or should we leave it just on the power board page?

Copying over the warning feels like it might be enough?

PeterJCLaw: > We currently don't mention on the brain board page that it needs to be L2, nor do we mention how to power it. Do we think this is worth doing, or should we leave it just on the power board page?
> 
> Copying over the warning feels like it might be enough?

This was discussed on the PR which updated that page (#355) and we concluded that it should be mentioned on the kit assembly page, as that's where this sort of thing has gone in the past. That page does mention this requirement and the fact that it needs a CamCon based cable (#369).

WillB97: I think it's worth saying that it's powered from 12 volts to avoid confusion.

WillB97:
We also need to update the available ports on https://studentrobotics.org/docs/programming/sr/power/

kit/brain_board/index.md
sedders123: [79f1ece](https://github.com/srobo/docs/pull/375/commits/79f1ece33799ec29e40577ab284230cd65f0a1e4)

8BitJosh (approved):

PR #381, opened 2022-10-21
Update kit assembly video

### LINKED ISSUE #380: Update kit assembly video

sedders123:
Closes #380

image

### DIFF
--- a/kit/assembly.md
+++ b/kit/assembly.md
@@ -65,7 +65,7 @@ The following video contains an overview which covers the [brain board](/docs/ki
 <iframe class="video center"
         height="315"
         width="560"
-        src="https://www.youtube-nocookie.com/embed/vhZjjCdhVE0?rel=0"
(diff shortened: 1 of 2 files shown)


kit/assembly.md
```diff
@@ -62,10 +62,12 @@ Video
...
-        height="315"
-        width="560"
-        src="https://www.youtube-nocookie.com/embed/vhZjjCdhVE0?rel=0"
-        frameborder="0"
-        allowfullscreen>
-</iframe>
+<div class="centered-content">
+  <iframe class="video center"
+          height="315"
+          width="560"
```
PeterJCLaw, line 68: This is much more square than the video actually is, is that intentional? (I realise it was like this before, no idea why)

PeterJCLaw: Scratch that, the extra height is coming from CSS. Shall we drop the extra CSS rule (just the bit which messes with the height)?

PeterJCLaw (approved):

PR #383, opened 2022-10-21
Add docs on the robot-settings.toml file

trickeydan:
We need to document how the `robot-settings.toml` file is created and what the settings in it mean.

This is the file that students need to edit to customise their WiFi password. It is automatically generated if a USB drive is inserted containing a `robot.py`

### DIFF
--- a/kit/brain_board/index.md
+++ b/kit/brain_board/index.md
@@ -78,3 +78,19 @@ If you choose to use a tool other than Etcher, you may need to extract the `srob
    Your computer may complain that the SD card is no longer readable, however
    this is expected as the data being written to the SD card is not in a format
    that either Windows or macOS can natively understand.
+
+## Robot Settings
+
+Some of the features on your robot are configured using a settings file, called `robot-settings.toml`. This file is automatically created on your USB drive the first time that you run any code on your robot.
+
+You can edit the settings file using your IDE or any text editor.
+
+The robot settings file contains the following settings:
+
+| Setting Name          | Description                                | Default Value      |
+|-----------------------|--------------------------------------------|--------------------|
+| `team_tla`            | Three Letter Acryonym (TLA) for your team. | Randomly generated |
+| `wifi_psk`            | Password for the Robot WiFi                | Randomly generated |
+| `wifi_region`         | Region Identifier for the WiFi             | `GB`               |
+| `wifi_enabled`        | Enables the WiFi                           | `true`             |
+| `usercode_entrypoint` | The entry point to your Python code        | `robot.py`         |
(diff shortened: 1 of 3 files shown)


kit/brain_board/index.md
```diff
@@ -78,3 +78,19 @@ If you choose to use a tool other than Etcher, you may need to extract the `srob
    Your computer may complain that the SD card is no longer readable, however
    this is expected as the data being written to the SD card is not in a format
    that either Windows or macOS can natively understand.
+
+## Robot Settings
+
+Some of the features on your robot are configured using a settings file, called `robot-settings.toml`. This file is automatically created on your USB drive the first time that you run any code on your robot.
+
+You can edit the settings file using your IDE or any text editor.
```
RealOrangeOne, line 86: Question: Kinda unrelated to this PR, but does the file use CRLF endings? Wondering how easy the file will be to edit on Windows.

RealOrangeOne (approved):

kit/brain_board/index.md
```diff
@@ -78,3 +78,19 @@ If you choose to use a tool other than Etcher, you may need to extract the `srob
...
+
+The robot settings file contains the following settings:
+
+| Setting Name          | Description                                | Default Value      |
+|-----------------------|--------------------------------------------|--------------------|
+| `team_tla`            | Three Letter Acryonym (TLA) for your team. | Randomly generated |
+| `wifi_psk`            | Password for the Robot WiFi                | Randomly generated |
+| `wifi_region`         | Region Identifier for the WiFi             | `GB`               |
+| `wifi_enabled`        | Enables the WiFi                           | `True`             |
```
raccube, line 95: nit: seems to be written out as `true` rather than `True`, off the top of my head, very likely that it takes both.
```suggestion
| `wifi_enabled`        | Enables the WiFi                           | `true`             |
```

raccube (approved):

kit/brain_board/index.md
RealOrangeOne: TOML booleans are in fact lowercase: https://toml.io/en/v1.0.0#boolean

kit/brain_board/index.md
raccube: Windows Notepad supports Unix line endings since Windows 10 21H1.

trickeydan: Oh good point actually... I suspect it generates without them but will read the file back fine if they are present. Opened issue on Astoria to track

kit/brain_board/index.md
trickeydan: We're using pydantic to parse these files so it should be pretty lenient on what it accepts. Updated and force pushed.

sedders123 (approved):

RealOrangeOne (approved):

PR #386, opened 2022-10-21
Add section on purple light of death to the Interactive Troubleshooter

sedders123:


resources/troubleshooter/data.json
```diff
@@ -407,6 +407,21 @@
                 }
             ]
         },
+        "servo-plod": {
+            "question": "Is there a purple LED on the Servo Board?",
```
trickeydan, line 411:
```suggestion
            "question": "Is there a pink LED on the Servo Board?",
```

I'd say it's more of a pink light, the red LED is much brighter than the blue

resources/troubleshooter/data.json
```diff
@@ -407,6 +407,21 @@
                 }
             ]
         },
+        "servo-plod": {
+            "question": "Is there a pink LED on the Servo Board?",
```
sedders123, line 411:
```suggestion
            "question": "Is there a pink light on the Servo Board?",
```

trickeydan (approved):

PR #390, opened 2022-10-21
Add markers for SR2023

sedders123:


WillB97 (approved):

PeterJCLaw:
@sedders123 are you able to add the command line which was used to generate these markers? It would be great to have that for posterity and/or if we need to regenerate them.

WillB97:
https://github.com/srobo/game-markers/blob/master/SR2023/sr2023-markers

PR #402, opened 2022-10-24
Fix marker size description

RealOrangeOne:
A markers' size is actually the size of the marker itself (the black bit) rather than the grey border as it was with Koki.

### DIFF
--- a/programming/sr/vision/markers.md
+++ b/programming/sr/vision/markers.md
@@ -21,7 +21,7 @@ Details of the types and size of markers used in the game can be found in the [r
 You can download all the markers as a single [ZIP file](/docs/resources/2023/sr-markers-sr2023.zip) or individually from the [git repo](https://github.com/srobo/game-markers/tree/master/SR2023/markers).
 
 You must ensure that your PDF viewer is not resizing the documents when printing.
-This can be checked by measuring the grey box around the marker and comparing this to the size defined in the rules.
+This can be checked by measuring the black marker and comparing this to the size defined in the rules.
 If the printed marker is not the correct size then the distance information reported by the vision API will be wrong.
 
 Note that a different set of markers will be used in the arenas at the competition.
--- a/programming/sr/vision/markers.md
+++ b/programming/sr/vision/markers.md
@@ -28,6 +28,6 @@ Note that a different set of markers will be used in the arenas at the competiti
 However, this is not something you need to worry about.
 We will handle this for you automatically when your robot is started in competition mode.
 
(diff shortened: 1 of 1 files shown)


programming/sr/vision/markers.md
```diff
@@ -21,13 +21,13 @@ Details of the types and size of markers used in the game can be found in the [r
 You can download all the markers as a single [ZIP file](/docs/resources/2023/sr-markers-sr2023.zip) or individually from the [git repo](https://github.com/srobo/game-markers/tree/master/SR2023/markers).
 
 You must ensure that your PDF viewer is not resizing the documents when printing.
-This can be checked by measuring the grey box around the marker and comparing this to the size defined in the rules.
+This can be checked by measuring the black marker and comparing this to the size defined in the rules.
```
PeterJCLaw, line 24: For my understanding: do AprilTags markers always have at least one square's worth of black all the way around the outside of the marker?

programming/sr/vision/markers.md
```diff
@@ -21,13 +21,13 @@ Details of the types and size of markers used in the game can be found in the [r
...
 If the printed marker is not the correct size then the distance information reported by the vision API will be wrong.
 
 Note that a different set of markers will be used in the arenas at the competition.
 However, this is not something you need to worry about.
 We will handle this for you automatically when your robot is started in competition mode.
 
-The white space around the markers is very important -- without it, the markers probably won't be recognised. This white border is needed to ensure there's enough contrast between the black marker and whatever it's attached to - its size isn't strictly important.
+The white space around the markers is very important -- without it, the markers probably won't be recognised. This white border is needed to ensure there's enough contrast between the black marker and whatever it's attached to - its size isn't strictly important. The grey border around the marker is the same size as 1 pixel in the marker, which is a good target for contrast.
```
PeterJCLaw, line 31: Does this mean the width of the grey line itself or the offset of the line from the black edge is one pixel?

programming/sr/vision/markers.md
RealOrangeOne: Yes, that's correct. We then add an extra bit of padding (coincidentally 1 square) to help contrast between the backing surface and marker.

programming/sr/vision/markers.md
RealOrangeOne: Ah yes that wording could be improved. The grey line isn't 1 pixel, it's the space between the line and marker which is 1 pixel.

PeterJCLaw:
```suggestion
The white space around the markers is very important -- without it, the markers probably won't be recognised. This white border is needed to ensure there's enough contrast between the black marker and whatever it's attached to - its size isn't strictly important. The white space around the marker is the same size as one block within the marker, which is a good target for contrast.
```

PeterJCLaw (approved):

PR #408, opened 2022-11-19
Only run the Docker build when relevant changes are made

PeterJCLaw:
The Docker build is comparatively slow (1-2 minutes) and mostly redundant given that the majority of PRs change docs content rather than configuration or build system.


RealOrangeOne:
I tend to find these kinds of "Only run CI when a certain file changes" quite brittle. Is the fact CI takes a little longer a problem? Generally getting a PR reviewed in under a minute is quite difficult, and now we live in a world with auto-merge, it shouldn't be too restrictive.

PeterJCLaw:
It's not about the review itself; it's more that when putting up a draft PR waiting for CI to go green before marking ready for review means waiting a long time.

PeterJCLaw:
Yes, I know I can/should run the checks locally, however if we want to encourage contributions from a wider audience then supporting the develop-in-GitHub case and making that have a fast feedback loop is worthwhile.

PeterJCLaw:
@RealOrangeOne I appreciate the concern around fragility, however looking at the time since this PR was created there has been one change that it's needed updating to accommodate and that change wasn't related to Docker (it was the renaming of `master` to `main`).

It would be great if we could ship this so that contributors aren't slowed down by this mostly redundant CI.

PR #411, opened 2022-11-22
Document firewall port range

trickeydan:


kit/brain_board/index.md
```diff
@@ -99,3 +99,18 @@ The robot settings file contains the following settings:
...
+
+## Firewall
+
+<div class="info">
+This documentation refers to a feature which is only available from software version `2023.1.0` and later.
+</div>
+
+The Brain Board has a [network firewall](https://en.wikipedia.org/wiki/Firewall_(computing)) that prevents access to applications and services on the Brain Board unless it is required.
+
+If you are running code that uses networking, such as `Flask`, you will need to use a port from the range that is allocated for competitor use:
```
RealOrangeOne, line 111: Suggestion: Given we don't ship flask, it might not be a helpful example.
```suggestion
If you are running code that uses networking, you will need to use a port from the range that is allocated for competitor use:
```

RealOrangeOne (approved):
Can confirm this matches our config: https://github.com/srobo/robot-image/blob/main/sources/meta-srobo/recipes-robot/srobo-firewall/srobo-firewall/nftables.conf#L25

raccube (approved):

PR #413, opened 2022-11-23
Add docs for remote debugging with VS Code

raccube:


PeterJCLaw:
If I'm reading this right it seems that the debugger is always active within the usercode process. Is that right? I would have expected there to be a performance (and security, though that's probably not a concern here) implication if so -- has that been considered?

raccube:
Ah. I completely forgot, there's a snippet that needs to be added.

programming/editors/vscode.md
```diff
@@ -35,3 +35,57 @@ you'll need to make the library available within the Python environment that you
...
+## Remote Debugging
+
+<div class="info">
+This documentation refers to a feature which is only available from software version `2023.1.0` and later.
+</div>
+
+When connected to the [robot's WiFi hotspot](/docs/kit/wifi), it is possible to attach VS Code's
+debugger to the robot by performing the following steps:
+
+1. Ensure the [Python extension](#python-extension) is installed
```
PeterJCLaw, line 48: nit (also below): it looks like the prevailing style is to treat each bullet in the list as a sentence and thus include a full stop after each (if there isn't other punctuation).

PeterJCLaw (approved):
Some small suggestions inline and I've not got a way to test this with a kit, however this looks right from my experience with VSCode & debugpy.

programming/editors/vscode.md
```diff
@@ -35,3 +35,57 @@ you'll need to make the library available within the Python environment that you
...
+            ],
+            "justMyCode": true
+        }
+    ]
+}
+```
+5. Add the following snippet to the top of your code:
+```python
+import debugpy
+debugpy.listen(("0.0.0.0", 5678))
```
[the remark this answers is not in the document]
raccube: Yep.

programming/editors/vscode.md
raccube: 277dbb8

programming/editors/vscode.md
```diff
@@ -35,3 +35,57 @@ you'll need to make the library available within the Python environment that you
 3. Run `pip install sr.robot3`.
 
 [code-completion]: https://en.wikipedia.org/wiki/Autocomplete#In_source_code_editors
+
+## Remote Debugging
+
+<div class="info">
+This documentation refers to a feature which is only available from software version `2023.1.0` and later.
```
trickeydan, line 42:
```suggestion
This documentation refers to a feature which is only available from software version <code>2023.1.0</code> and later.
```

Backticks don't render inside HTML

programming/editors/vscode.md
```diff
@@ -35,3 +35,57 @@ you'll need to make the library available within the Python environment that you
...
+</div>
+
+When connected to the [robot's WiFi hotspot](/docs/kit/wifi), it is possible to attach VS Code's
```
trickeydan, line 45:
```suggestion
When connected to the [robot's WiFi hotspot]({{ site.baseurl }}/kit/wifi), it is possible to attach VS Code's
```

programming/editors/vscode.md
PeterJCLaw: Good spot. An alternative is to add `markdown="1"` to the containing HTML element.

PR #422, opened 2022-12-14
Specify the mass of the Bronze tokens

sedders123:
Use the SR2023.4 rulebook


PeterJCLaw:
Unable to approve as the file does not match the one in CI. The diff is hard to read, so re-created as #423 as discussed.

PR #427, opened 2023-01-07
Remove references to Zoloto

trickeydan:
Zoloto is no longer used.

### DIFF
--- a/.spelling
+++ b/.spelling
@@ -25,7 +25,6 @@ studentrobotics
 usbkey
 usercode
 Webots
-Zoloto
 
 # Markup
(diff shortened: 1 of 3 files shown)


PeterJCLaw:
Should we link to the library which has replaced it?

trickeydan:
> Should we link to the library which has replaced it?

I don't think we should yet. Once @WillB97 has transferred the repository to `srobo` and we have decided what level of documentation needs to live with the repo, then I think we can revisit it. I think for now, just linking to AprilTag is enough.

PeterJCLaw (approved):

PR #433, opened 2023-01-21
Update libraries for the 2023.2.1 image

WillB97:


trickeydan (approved):
Thanks Will 

We should work on automating this a bit more. Maybe a robot.py that spits out markdown?

PR #437, opened 2023-02-02
Update the vision docs to match reality

PeterJCLaw:
Even before the changes in the Dec 1st kit update these docs were wildly different to how the vision system actually behaved. This update brings these to parity with the default behaviour of the kit (and soon the simulator).

Ideally we would have some diagrams to show how the axes work, however the diagrams we have are substantially incorrect and will need replacing. Dropping them clarifies that work is needed before they can be brought back (though recreating them from scratch might be easier as we don't seem to have the source files).

Edit: the intended scope here is to update the kit vision API types. I've deliberately ignored the changes needed for the upcoming simulator changes.


programming/sr/vision/index.md
```diff
@@ -171,62 +184,71 @@ One (possibly both) of them may change to resolve this.
...
 
-roll
+rot_z
 :   Rotation of the marker about the cartesian z-axis, in radians.
 
     Turning a marker anticlockwise (as viewed from the camera) increases the
-    value of `rot_z`, while turning it clockwise decreases it. A value of 0
-    indicates that the marker is upright.
+    value of `rot_z`, while turning it clockwise decreases it. A value of π
+    or -π indicates that the marker is upright.
```
WillB97, line 206: This should be 0 when the marker is upright while using zoloto markers.

programming/sr/vision/index.md
```diff
@@ -171,62 +184,71 @@ One (possibly both) of them may change to resolve this.
...
```
WillB97, line 207:
```suggestion

There are additional attributes for the [principal axis rotations of the marker](https://en.wikipedia.org/wiki/Aircraft_principal_axes).

```

programming/sr/vision/index.md
```diff
@@ -171,62 +184,71 @@ One (possibly both) of them may change to resolve this.
...
 
-rot_x
-:   An alias for `pitch`.
+roll
+:   An alias for `rot_x`.
 
-rot_y
-:   An alias for `yaw`.
+pitch
+:   An alias for `rot_y`.
```
WillB97, line 212:
```suggestion
:   A rotation about the transverse axis, in radians (An axis right to left across the token). Tilting the marker backward increases the value of `pitch`,
    while tilting it forwards decreases it. A value of 0 indicates that the marker is facing the camera square-on.
    
    This differs from `rot_x` in the zero point and direction to increase the value.
```

programming/sr/vision/index.md
```diff
@@ -171,62 +184,71 @@ One (possibly both) of them may change to resolve this.
...
 
-rot_z
-:   An alias for `roll`.
+yaw
+:   An alias for `rot_z`.
```
WillB97, line 215:
```suggestion
:   A rotation about the about the vertical axis, in radians (An axis top to bottom through the token).
    Turning a marker clockwise (as viewed from above) increases the value of
    `yaw`, while turning it anticlockwise decreases it. A value of 0 means
    that the marker is perpendicular to the line of sight of the camera.
    
    This differs from `rot_y` in the direction that increases the value.
```

programming/sr/vision/index.md
```diff
@@ -171,62 +184,71 @@ One (possibly both) of them may change to resolve this.
...
 The spherical coordinates system has three values to specify a specific point in space.
 
-![A visual representation of Spherical coordinates. Source: SourceBots]({{ site.baseurl }}/images/content/vision/spherical.png)
+distance
+:   The radial distance, the distance from the origin to the point, in millimetres.
+
+theta
+:   This is the angle from directly in front of the camera to the vector which
+    points to the location in the horizontal plane. A positive value indicates a
+    counter-clockwise rotation. Zero is at the centre of the image.
```
WillB97, line 236: In radians.

programming/sr/vision/index.md
```diff
@@ -146,20 +146,33 @@ orientation
 :   An [`Orientation`](#Orientation) instance describing the orientation of the marker.
 
 spherical
-:   A [`Spherical`](#Spherical) instance describing the position relative to the camera.
+:   A [`SphericalCoordinate`](#SphericalCoordinate) instance describing the position relative to the camera.
 
 cartesian
-:   A [`ThreeDCoordinate`](#ThreeDCoordinate) instance describing the absolute position of the marker relative to the camera.
+:   A [`CartesianCoordinates`](#CartesianCoordinates) instance describing the absolute position of the marker relative to the camera.
```
WillB97, line 152:
```suggestion
:   A [`CartesianCoordinates`](#CartesianCoordinates) instance describing the position of the marker relative to the camera.
```

programming/sr/vision/index.md
```diff
@@ -146,20 +146,33 @@ orientation
...
+
+[`PixelCoordinates`](#PixelCoordinates) {#PixelCoordinates}
 ---------
 
-A `Coordinate` object contains an `x` and `y` attribute. The exact meaning and unit of these attributes depends on its source.
+A named tuple of `x` and `y` coordinates for the point, in pixels relative to the top left of the image.
+
+~~~~~ python
+print(marker.pixel_centre.x, marker.pixel_centre.y)
+~~~~~
```
WillB97, line 163: Can we add some context around what this code block does.

programming/sr/vision/index.md
```diff
@@ -146,20 +146,33 @@ orientation
...
+[`CartesianCoordinates`](#CartesianCoordinates) {#CartesianCoordinates}
 ---------
 
-A `ThreeDCoordinate` object contains an `x`, `y` and `z` attribute. The exact meaning and unit of these attributes depends on its source.
+A named tuple of `x`, `y` and `z` coordinates for the point, in millimeters relative to the camera.
+Increasing values are to the right, below and away from the camera respectively.
+
+~~~~~ python
+print(marker.cartesian.x, marker.cartesian.y, marker.cartesian.z)
+~~~~~
```
WillB97, line 175: Can we add some context around what this code block does.

WillB97:
We should also state the differences in the API in the simulator, mainly:
- There is not a `R.camera.capture()` method.
- None of the camera methods support a `frame` argument as the recogniion system cannot be given arbitrary frames.
- There is not a `R.camera.save()` method.
- The quaternion returned by `Marker.orientation.quaternion` is only a tuple of the quaternion's w, x, y, z components.

programming/sr/vision/index.md
```diff
@@ -171,62 +184,71 @@ One (possibly both) of them may change to resolve this.
 
 An `Orientation` object describes the orientation of a marker.
 
-![A visual representation of how the orientation axes work. Source: SourceBots]({{ site.baseurl }}/images/content/vision/yaw-pitch-roll.png)
-
-pitch
+rot_x
 :   Rotation of the marker about the cartesian x-axis, in radians.
```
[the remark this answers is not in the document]
PeterJCLaw: Indeed it does. The axes are defined separately, above, and are the same for all of the vision objects. We don't repeat that information in the other types either, which I think is fine.

programming/sr/vision/index.md
PeterJCLaw: I'll double check but I tested these when I wrote the docs, so I think this is accurate.

programming/sr/vision/index.md
PeterJCLaw: This implies that it is conceptually more or different than `rot_y`, which it is not. This property is merely an alias for the same quantity.

PeterJCLaw:
> We should also state the differences in the API in the simulator, mainly:

I've updated the description to clarify that I'd deliberately not addressed the simulator differences here; to cover them in detail is going to need non-trivial additional changes which I'd rather address separately. (I'm also conscious that we haven't addressed the orientation issues in the simulator yet, so they may yet change anyway).

programming/sr/vision/index.md
WillB97: `pitch` is __not__ an alias of `rot_y` unless you are using april_vision, in zoloto mode the axis and zero-points of yaw, pitch and roll differ from the axis rotations.

PeterJCLaw: Both `Orientation.rot_y` and `Orientation.pitch` are implemented as `return self._yaw_pitch_roll[1]`, with no accounting for the mode of operation. They are thus identical. While the implementation may not be that one is strictly an alias of the other, that is functionally what they are.

Perhaps you're thinking of the spherical coordinates?

WillB97: `Orientation.rot_y` is implemented as `return self.yaw_pitch_roll[1]`, `Orientation.pitch` is implemented as `return self._yaw_pitch_roll[1]` (note the leading underscore). `_yaw_pitch_roll` is built from the quaternion that has not had the zoloto rotation applied in the init.

PeterJCLaw: Eugh, I didn't spot that. That's remarkably unclear code!

Will update this section with the details of the yaw/pitch/roll properties.

programming/sr/vision/index.md
PeterJCLaw: Is it intentional that this defines the zero point as "perpendicular to the line of sight of the camera" where others frame it as "facing the camera square-on"? Specifically: is there a semantic difference that this is trying to convey?
If not, I think we should use the same framing everywhere.

programming/sr/vision/index.md
PeterJCLaw: 153a2ef

programming/sr/vision/index.md
PeterJCLaw: 7e1b39f

programming/sr/vision/index.md
PeterJCLaw: 7e1b39f

programming/sr/vision/index.md
```diff
@@ -171,62 +184,71 @@ One (possibly both) of them may change to resolve this.
...
 
-    Leaning a marker away from the camera increases the value of `rot_x`, while
-    leaning it towards the camera decreases it. A value of 0 indicates that the
-    marker is upright.
+    Leaning a marker towards the camera increases the value of `rot_x`, while
+    leaning it away from the camera decreases it. A value of π or -π indicates
+    that the marker is upright.
```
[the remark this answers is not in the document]
PeterJCLaw: a287261

programming/sr/vision/index.md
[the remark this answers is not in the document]
WillB97: Stating "rot_x is a rotation about the x-axis" is a pointless statement as that is readily implied. Stating this is a pitch-like rotation allows them to relate it to something familar.

programming/sr/vision/index.md
PeterJCLaw: edecab2 :disappointed:

programming/sr/vision/index.md
WillB97: Face-on to the camera implies that the yaw is calculated relative to the tagent of a sphere aroud the camera while it is actually calculated relative to a flat plane in front of the camera.

PeterJCLaw: Ok, can we use the "perpendicular to the line of sight of the camera" for the pitch case too then? Presumably that suffers the same issue, just in a different plane.

WillB97: Yes, this is true for all bits talking about orientation.

programming/sr/vision/index.md
PeterJCLaw: 6de4868

programming/sr/vision/index.md
PeterJCLaw: 6de4868

programming/sr/vision/index.md
PeterJCLaw: 6de4868

programming/sr/vision/index.md
```diff
@@ -171,62 +184,71 @@ One (possibly both) of them may change to resolve this.
...
 
-    Turning a marker clockwise (as viewed from above) increases the value of
-    `rot_y`, while turning it anticlockwise decreases it. A value of 0 means
+    Turning a marker clockwise (as viewed from above) decreases the value of
+    `rot_y`, while turning it anticlockwise increases it. A value of 0 means
     that the marker is perpendicular to the line of sight of the camera.
 
-roll
+rot_z
 :   Rotation of the marker about the cartesian z-axis, in radians.
```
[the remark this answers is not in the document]
PeterJCLaw: 6de4868

programming/sr/vision/index.md
```diff
@@ -171,62 +184,71 @@ One (possibly both) of them may change to resolve this.
...
-    Leaning a marker away from the camera increases the value of `rot_x`, while
-    leaning it towards the camera decreases it. A value of 0 indicates that the
-    marker is upright.
+    Leaning a marker towards the camera increases the value of `rot_x`, while
+    leaning it away from the camera decreases it. A value of π or -π indicates
+    that the marker is upright.
 
-yaw
+rot_y
 :   Rotation of the marker about the cartesian y-axis, in radians.
```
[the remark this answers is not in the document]
PeterJCLaw: 6de4868

programming/sr/vision/index.md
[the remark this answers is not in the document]
PeterJCLaw: 6de4868

programming/sr/vision/index.md
```diff
@@ -171,62 +186,98 @@ One (possibly both) of them may change to resolve this.
...
+Finally there are attributes which express the orientation in other forms:
 
 rotation_matrix
-:   The rotation matrix represented by this orientation. A list of 3 lists, each with 3 items.
+:   The [rotation matrix](https://en.wikipedia.org/wiki/Rotation_matrix) represented by this orientation.
+    This 3×3 matrix is represented by a list of 3 lists, each with 3 values, in an arrangement compatible with tools such as `numpy`.
 
 quaternion
-:   The [Quarternion](https://kieranwynn.github.io/pyquaternion/#quaternion-features) instance represented by this orientation.
+:   The [quarternion](https://en.wikipedia.org/wiki/Quaternion) represented by this orientation.
```
WillB97, line 249:
```suggestion
:   The [quaternion](https://en.wikipedia.org/wiki/Quaternion) represented by this orientation.
```

programming/sr/vision/index.md
```diff
@@ -146,20 +146,35 @@ orientation
 :   An [`Orientation`](#Orientation) instance describing the orientation of the marker.
 
 spherical
-:   A [`Spherical`](#Spherical) instance describing the position relative to the camera.
+:   A [`SphericalCoordinate`](#SphericalCoordinate) instance describing the position relative to the camera.
 
 cartesian
-:   A [`ThreeDCoordinate`](#ThreeDCoordinate) instance describing the absolute position of the marker relative to the camera.
+:   A [`CartesianCoordinates`](#CartesianCoordinates) instance describing the position of the marker relative to the camera.
```
WillB97, line 152: Can we use the same wording for both here.

programming/sr/vision/index.md
```diff
@@ -171,62 +186,98 @@ One (possibly both) of them may change to resolve this.
...
+yaw
+:   A rotation about the about the vertical axis, in radians (an axis top to
+    bottom through the token). Turning a marker clockwise (as viewed from above)
+    increases the value of `yaw`, while turning it anticlockwise decreases it. A
+    value of 0 means that the marker is perpendicular to the line of sight of
+    the camera.
 
-rot_z
-:   An alias for `roll`.
+    This differs from `rot_y` in the direction that increases the value.
```
WillB97, line 223: Markdown looks to be converting this paragraph into a code block. Same for the extra paragraphs on the two below.

programming/sr/vision/index.md
PeterJCLaw: 3dc777c31130bd5cf271020a38d774cb069438a1

programming/sr/vision/index.md
WillB97: Screenshot 2023-02-08 at 22 34 30

PeterJCLaw: I think that's a function of GitHub's rendering. Here's what it looks like under Jekyll:
image

WillB97 (approved):

PR #438, opened 2023-02-04
Make webots version constraint more prominent

RealOrangeOne:
This makes it clearer to teams which versions we support, hopefully reducing compatibility issues


PeterJCLaw:
Or we could do a release which supports 23a?

RealOrangeOne:
That's also possible, but until then this is still correct and valid.

PR #454, opened 2023-04-11
RFC: Add search

RealOrangeOne:
Peek 2023-04-11 20-13

Add a site search for finding some things more easily. It's not perfect, but it's something.

Update: I realised,whilst writing this that this won't actually work, as we're still relying on GitHub Pages automatically deploying our site. If we move to manually deploying the site ourselves, it will then work. Until we have that, this is here for review.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -17,15 +17,13 @@ task :dependencies do
     sh('bundle install --path gems')
   end
 
+  sh('npm install')
+
   # Fix pathutil on Ruby 3; works around https://github.com/envygeeks/pathutil/pull/5
   # as suggested by https://stackoverflow.com/a/73909894/67873
   pathutil_path = `bundle exec gem which pathutil`.strip()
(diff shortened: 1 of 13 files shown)


package.json
```diff
@@ -1,13 +1,11 @@
...
   },
-  "@comments": [
-    "These dependencies are only for testing the site.",
-    "The site is built with Jekyll so site dependencies should be in the Gemfile."
-  ],
   "dependencies": {
-    "cspell": "5.18.5"
+    "cspell": "5.18.5",
+    "pagefind": "^0.12.0"
   }
```
PeterJCLaw, line 10: If we're moving to actually building JS stuff here we should move `cspell` into `devDependencies`.

Rakefile
PeterJCLaw, line 1: Is there a missing call to `npm build-search-index` in this file?

PeterJCLaw:
Is this related to the experimentation that was previously being done with Algolia?

PR #457, opened 2023-05-14
Suggest browsers preconnect to third party domains we use

PeterJCLaw:


RealOrangeOne (commented):
Looks like we also use CDNJS for font-awesome

PeterJCLaw:
> Looks like we also use CDNJS for font-awesome

We do, however that's already directly included in the same file so there's unlikely to be any benefit. The origins mentioned here are deliberately only those which are not directly mentioned in the initial HTML file -- so that browsers can start to connect to them without waiting for the intermediate resource to download.

raccube (approved):

PR #459, opened 2023-05-24
Move the SR2023 rules to the archive

PeterJCLaw:
The arena image was constructed by manually adjusting the diagram from the rules to remove the dimensions and other artefacts, then exporting as PNG via inkscape.

Where previously we've replaced the content of the rules index page entirely at this point in the year, it feels like having a conditional (even a constant one) will make this process slightly simpler in future.


images/content/rules/arena-2023.png
RealOrangeOne, line 1: Issue: This PNG render has missed the horizontal zone boundaries.

rules/index.md
```diff
@@ -1,17 +1,21 @@
...
 ---
 
-2023 Rules
+2024 Rules
 ==========
 
-[<img class="left" src="{{ site.baseurl }}/resources/2023/rulebook.png" />]({{ site.baseurl }}/resources/2023/rulebook.html)
-The rules, regulations and specifications for the SR2023 competition can be
-found in the [rulebook]({{ site.baseurl }}/resources/2023/rulebook.html).
+{% if false %}
```
RealOrangeOne, line 9: I'm not sure I agree with this being useful. All the content needs to change every time we switch the conditional anyway.

Perhaps a better implementation would be to use an `include` template for each possible value, given the text is almost always the same:

```html+jinja
{% include "rules-link.html" year=2024 %}

{% include "rules-pending.html" year=2024 %}
```

This way, we can make sure we keep the same wording between years, and switching between becomes as simple as commenting / uncommenting template (with perhaps updating a year value)

Perhaps even pulling the year from some frontmatter so it's only defined once, both for passing into the templates and for the page title.

RealOrangeOne (changes requested):

images/content/rules/arena-2023.png
PeterJCLaw: Would be nice if the SVG had used layers for this...

PeterJCLaw: https://github.com/srobo/docs/pull/459/commits/624e535b850b845f97669176ae8aef99c9438386

rules/index.md
RealOrangeOne: Perhaps the solution to improve readability is to instead drive it from a frontmatter variable.

```
---
show_rulebook: true
---

{% if page.show_rulebook %}
...
{% else %}
...
{% endif%}
```

It's _almost_ the same, but makes it far more readable as to what's going on, without someone thinking "`if false` will never happen, let's clean up that dead code".

PeterJCLaw: Ah, neat. I hadn't thought of/wasn't aware that you could do that. Will do that.

PeterJCLaw: 8a774f7ba89b58e1e2f867e9da087b4f2aa55db9

RealOrangeOne (approved):

PR #506, opened 2023-09-09
Update page on python libraries

8BitJosh:
Libraries synced between the image and simulator, python libs page updated to reflect this

Simulator libraries https://github.com/srobo/competition-simulator/blob/main/libraries.txt
Robot image libraries https://github.com/srobo/robot-image/blob/main/libraries.txt
Robot image requirements https://github.com/srobo/robot-image/blob/main/requirements.txt

PR #507, opened 2023-09-09
Update Python & Webots versions the simulator uses

PeterJCLaw:
This deliberately breaks the link to the simulation files as we have not yet released a version for SR2024, so any link would be incorrect relative to the documented versions.

See also #506.


8BitJosh (approved):

8BitJosh:
Can we get this merged, I want to restructure this section of the docs and merging this first will save some headache later

PR #512, opened 2023-09-13
Document battery charger barrel size

RealOrangeOne:
To be confirmed this weekend.


8BitJosh:
Can we change this to 5.5x2.5mm

8BitJosh (approved):

PR #540, opened 2023-10-13
Refactor the simulator docs

PeterJCLaw:
This drops the idea of programming in the simulator from these docs, since we don't really document that here. Instead that is covered by the main programming docs when read alongside these.

That page's content moves to the docs on using the simulator and (mostly) to a new page about the simulated robot. What was two sections on simulated time becomes essentially one more cohesive section.

The index page is also considerably fleshed out to provide navigation between these new sections as well as a clearer overview.


Adimote (approved):
Thanks for doing this! I agree the previous layout was a bit weird, but I wasn't sure what a better solution would be

PR #545, opened 2023-10-14
Replace broken PNG

### LINKED ISSUE #544: Broken marker PNG
The image at https://studentrobotics.org/docs/programming/vision/markers.html in the top right (https://studentrobotics.org/docs/images/content/vision/marker-0.png) appears to be broken due to some chunks having the wrong resolution.

It seems to actually display ok in chrome, but not in Firefox. TinyPNG can't decode it.
https://www.nayuki.io/page/png-file-chunk-inspector can be useful to see the errors.

WillB97:
Fixes #544


PeterJCLaw (approved):
Thanks!
Checked via comparing to what Chromium renders.

raccube (approved):

PR #546, opened 2023-10-16
Add links to a version-zero release of the SR2024 simulator

PeterJCLaw:
Useful if we do want to do this to help teams get set up ahead of time.

Note: at the time of writing the link included here doesn't exist.

Depends on us shipping https://github.com/srobo/competition-simulator/pull/435 as `sr2024.0`.


PeterJCLaw:
We didn't do this in the end.

PR #569, opened 2023-10-22
Set the HTML language for better accessibility

PeterJCLaw:


raccube (approved):

raccube:
Since HTML 5 this is actually required.

PR #571, opened 2023-11-04
Drop interim rules page

RealOrangeOne:
This removes the unnecessary extra page with a link to the rules, making them easier to find.

This also means `srobo.org/rules` links to the rulebook directly.


WillB97 (approved):

RealOrangeOne:
The rules page now has 2 documents on, so this is no longer valid.

PR #577, opened 2024-01-20
Fix ultrasound units

RealOrangeOne:
It actually returns a voltage, wich needs mapping to a distance


WillB97 (approved):
Maybe we should do `(5/2)` rather than just using 2.5

Issue #488, opened 2023-08-28
Docker for rake validate


8BitJosh:
Add some sort of Docker based tool for validating (check links/spelling) the docs locally.

PeterJCLaw:
Could you expand on why Ruby/Rake are a pain on Windows? Is this a general issue in the ecosystem or something specific to how we're set up? Would it be useful for example for us to have more info on how to set them up and/or guidance on e.g: using them under WSL (assuming that that might be better than under plain Windows)?

There already is a Docker setup I think in the docs repo, though I don't use it. I had thought it already supported running the validate stuff. Personally I find its presence is a negative as it makes the CI slower. I'm on Linux (Ubuntu) though, and am able to run everything locally. It also therefore wouldn't occur to me to update it if things changed, which means it's liable to get out of sync.

For spell checking if you're using VSCode I'd encourage installing the Code Spell Checker extension. That should run the same spell checker in your editor as we run in CI. This is the main reason that we use the checker package that we do. I see however that that wasn't being suggested to users of the repo though, I've fixed that now (copied over from the wesite repo).
I don't know if it's supported in other editors, though it feels like it easily could be.

PeterJCLaw:
https://github.com/srobo/docs/pull/585 adds more information on how to run the checks locally.

WillB97:
That doesn't solve this problem though. This is about being able to use the validate tools with docker, not running them locally.

PeterJCLaw:
#585 attempts to address the reason for this request -- that setting these up is hard, which I've assumed is due to lack of information about what's needed. I've asked for further information about why it's hard but haven't had a response -- hence the guesswork.

PR #585, opened 2024-03-02
Various README improvements

PeterJCLaw:
This aims to make the process easier for people not already familiar with the setup, primarily by guiding them through the preferred (local install) path which has each of the necessary steps laid out. It also explains how to run the spell checking locally.

Finally it calls out the security risks associated with the Docker path, removes the instructions there to reduce the chances that people not equipped to judge those risks take that path and implies (correctly) that the spell checking is not supported under Docker.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,6 +26,8 @@ Once setup, the site will be accessible on http://localhost:4000/docs/
     $ rake dev
     ```
 
+4. View the website at <http://localhost:4000/docs/>
+
 ## Making changes
 
 When you've made a change, either push it to a forked repository, or to a
--- a/README.md
+++ b/README.md
@@ -4,12 +4,6 @@ The Student Robotics public documentation.
 
 ## Getting Started
 
-For ease of setup, a Docker container is provided. Simply install Docker and `docker-compose`, and run `docker-compose up`.
-
-Once setup, the site will be accessible on http://localhost:4000/docs/
-
-### Manual
-
 0. [Clone this repo][clone-repo]
 
 1. [Install Ruby][install-ruby]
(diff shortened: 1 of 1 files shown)


WillB97:
Is the local setup actually the preferred method? The primary change here seems to be about changing the preference from docker to local. 

Docker has the advantage that it doesn't require users to install various additional software (node, npm, rake) just to update the content. This is made more notable by the old version of node and rake needed in this repository.

PeterJCLaw:
> Is the local setup actually the preferred method? The primary change here seems to be about changing the preference from docker to local.

Yes, local setup is the method which is common to all our Jekyll sites. The Docker setup in this repo is an exception to that and one which was poorly documented. Note that e.g: the instructions for the Docker setup didn't previously note the need to clone the repo nor provide any information on how to set up or use Docker.

Since any non-trivial operations in this repo require the developer to understand the Ruby/Jekyll/Liquid setup to at least a basic level, we need to provide those instructions anyway. Any support for Docker is then a strict increase in the maintenance burden.

> Docker has the advantage that it doesn't require users to install various additional software (node, npm, rake) just to update the content. This is made more notable by the old version of node and rake needed in this repository.

Using Docker is not _better_ it is just _different_. It's not always easier to install nor is it even fewer things to install. Installing Docker is something which _requires_ the user to have local admin on their computer, as well as having an awareness of the security impacts of installing a container engine. The latter is especially pertinent given the current security issues with the Docker based setup here.

As noted previously -- since we need to support the Ruby route _anyway_, any issues which exist with that need fixing regardless of whether we also support the Docker route.

Please also bear in mind that whatever is done for this repo needs replicating for all our other Jekyll repos (of which there are four), so the incremental effort for having an extra setup method is not trivial.

Finally, as noted in other issues on this repo (e.g: #488) the current Docker setup is somewhat out of sync with the Ruby setup -- meaning that users still need to install the other dependencies anyway. I'd much rather put the effort into having a single approach we can make work well rather than two approaches which work badly.

PeterJCLaw:
Rebased to fix lots of conflicts.

PR #586, opened 2024-03-03
Lists start from one

WillB97:
People count from one, so should our lists.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@ Once setup, the site will be accessible on http://localhost:4000/docs/
 
 ### Manual
 
-0. [Clone this repo][clone-repo]
+1. [Clone this repo][clone-repo]
 
 1. [Install Ruby][install-ruby]
 


PeterJCLaw (changes requested):
These lists starting from 0 is intentional -- it's somewhat assumed that people may have already cloned the repo if they're reading the setup instructions.

WillB97:
> These lists starting from 0 is intentional -- it's somewhat assumed that people may have already cloned the repo if they're reading the setup instructions.

Then they just skip over the step it doesn't remove the fact that the step exists.

PeterJCLaw (approved):

RealOrangeOne:
If we're going to make this change, I'd suggest actually renumbering the markdown source so the numbers are sequential.

PeterJCLaw:
> If we're going to make this change, I'd suggest actually renumbering the markdown source so the numbers are sequential.

I'd actually suggest going the other way -- make them all `1.` so that there's no redundant diff when changes are made. https://github.com/srobo/docs/pull/585 actually does that as it was doing some other related tidying.

PR #600, opened 2024-08-03
Use a consistent mechanism to upload our reference build

PeterJCLaw:
Something changed recently in how actions/upload-artifact creates its archives, such that symlinked files behaved differently under that action than under actions/upload-pages-artifact. This broke our diffing logic.

Move to using the same upload approach for both the local and reference main builds, to avoid spurious errors.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -82,11 +82,10 @@ jobs:
       - name: Build
         run: rake build
 
-      - name: Upload build site
-        uses: actions/upload-artifact@v4
+      - name: Upload reference artifact
+        uses: actions/upload-pages-artifact@v3
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/publish.yml
```diff
@@ -113,7 +112,14 @@ jobs:
...
           name: build-from-main
-          path: main
+
+      - name: Unpack main build
+        run: |
+          mkdir main
+          pushd main
+          tar --extract --file ../artifact.tar
+          popd
+          mv artifact.tar main.tar
```
PeterJCLaw, line 122: This is copy/paste from the above logic for `local`, with the name swapped out.

RealOrangeOne (approved):

PR #608, opened 2024-08-31
Provide more information for supervisors

PeterJCLaw:
This extends the supervisor page and adds a new page about the overall programme structure. The goal here is to address the key aspects of https://github.com/srobo/tasks/issues/1357, though to fully complete that we'll also need to actually get these docs in front of supervisors early in the cycle.

This may also go some way towards other issues linked from that one, though I don't think fully addresses any single one of them. Again this is partly about getting the information in other places (though hopefully this is now something we can more easily/usefully link to).

As part of this change I've also linked to the team-supervisor ("Running a team") page from the signup form for SR2025. Ideally we'd include this change before that goes live, but even as it stands it's a useful resource for them.

Ideally I'd also like to have a separate page just about the competition event, but that can wait for another time. (Also I believe there's already a [draft somewhere](https://docs.google.com/document/d/1khsxcb84PZPm0071Vwkz4JfBY3z-xruAc9cs8UmUMSw/edit), which could usefully be finished)

### DIFF
--- a/robots_101/programme_structure.md
+++ b/robots_101/programme_structure.md
@@ -10,7 +10,7 @@ But what actually happens during the course of the competition year?
 
 ## Registration
 
-Registration for the competition usually opens in late summer. Teams usually
+Registration for the competition usually opens in early September. Teams usually
 consist of 4-8 competitors attending a school or college, though other groups
 are also free to enter. Your team must have a responsible adult who will oversee
 the team (a role we call the [team supervisor][team-supervisor]) and they must
--- a/robots_101/programme_structure.md
+++ b/robots_101/programme_structure.md
@@ -10,7 +10,7 @@ But what actually happens during the course of the competition year?
 
 ## Registration
 
-Registration for the competition usually open in late summer. Teams usually
+Registration for the competition usually opens in late summer. Teams usually
 consist of 4-8 competitors attending a school or college, though other groups
 are also free to enter. Your team must have a responsible adult who will oversee
 the team (a role we call the [team supervisor][team-supervisor]) and they must
--- a/robots_101/programme_structure.md
+++ b/robots_101/programme_structure.md
@@ -65,6 +65,14 @@ more direct help with your robot.
 We provide a space for you to work in, with power and internet access, as well
 as volunteers able to help with the kits and hands-on guidance with your robot.
 
+## Virtual Competition
+
+The Virtual Competition is often the first change you'll have to test your
+strategies against other teams. You'll submit your robot code for running
+against other teams in the [simulator][simulator]. The matches are livestreamed
+on our [YouTube Channel][youtube] and you'll get a copy of your robot's logs
+afterwards so you can continue to improve it.
+
 ## Competition
 
 The Competition event is a two day event in spring. You'll put your robot to the
@@ -88,3 +96,5 @@ afternoon you'll advance into the knockout stages, and can go on to win prizes.
 [kit]: {{ site.baseurl }}/kit/
 [discord]: {{ site.baseurl }}/tutorials/discord
 [tech-days]: {{ site.baseurl }}/robots_101/tech_days
+[simulator]: {{ site.baseurl }}/simulator/
+[youtube]: https://www.youtube.com/@studentrobotics
--- a/robots_101/programme_structure.md
+++ b/robots_101/programme_structure.md
@@ -75,8 +75,8 @@ afterwards so you can continue to improve it.
 
 ## Competition
 
-The Competition event is a two day event in spring. You'll put your robot to the
-test in a weekend of tinkering and matches.
+The Competition event is a two day event in spring. It's the ultimate test for
+your robot in a weekend of tinkering and matches.
 
 You'll have a team pit for you to work in, with power and internet access, and
 throughout the weekend there are volunteers around to help with the kits and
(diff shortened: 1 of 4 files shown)


robots_101/team_supervisor.md
```diff
@@ -17,7 +21,18 @@ You’ll be our point of contact with the team. If you have any questions during
 We aim to host our Kickstart event and Tech Days in multiple locations to make it more convenient for you to travel. However, you will still need to arrange to travel to these places.
 This is especially important to book for the competition, as you will likely need to arrange to stay overnight near the venue.
 
-Our kit includes a battery and a few boards to get your robot started. However, your team will need additional components and materials from which to build their robot. For the chassis, your team has several options. Teams often use cardboard, MDF, Aluminium, and/or Acrylic. In terms of electronic components, teams often make use of:
+## The team
```
PeterJCLaw, line 24: Not super happy with this section heading name; would appreciate some suggestions.

robots_101/programme_structure.md
```diff
@@ -0,0 +1,90 @@
...
+---
+
+# Robots 101 - Programme Structure
+
+Student Robotics is an annual robotics competition for people aged 16-19.
+But what actually happens during the course of the competition year?
+
+## Registration
+
+Registration for the competition usually open in late summer. Teams usually
```
antonnikitin97, line 13: Registration usually _opens_.

robots_101/programme_structure.md
```diff
@@ -0,0 +1,90 @@
...
+
+Throughout the year [Tech Days][tech-days] provide opportunities to spend a
+whole day (usually a Saturday) working on your robot, with lots of help
+available. They’re also an opportunity to see how other teams are doing or get
+more direct help with your robot.
+
+We provide a space for you to work in, with power and internet access, as well
+as volunteers able to help with the kits and hands-on guidance with your robot.
+
+## Competition
```
antonnikitin97, line 68: There's no mention of the virtual competition here - should there be? :)

robots_101/team_supervisor.md
```diff
@@ -7,17 +7,32 @@ title: Robots 101 - So you're running a team
...
+- Information about the [competition event]({{ site.baseurl }}/robots_101/programme_structure#competition) (usually in late March or April)
 
 We aim to host our Kickstart event and Tech Days in multiple locations to make it more convenient for you to travel. However, you will still need to arrange to travel to these places.
 This is especially important to book for the competition, as you will likely need to arrange to stay overnight near the venue.
 
-Our kit includes a battery and a few boards to get your robot started. However, your team will need additional components and materials from which to build their robot. For the chassis, your team has several options. Teams often use cardboard, MDF, Aluminium, and/or Acrylic. In terms of electronic components, teams often make use of:
+## The team
+
+Teams usually consist of 4-8 competitors aged 16-19 and are often attending a school or college.
+However, the team doesn't necessarily need to represent a school -- any group of 16-19 year-olds, with a responsible adult, can enter.
```
antonnikitin97, line 27: Is it worth defining what a `responsible adult` is here?

robots_101/programme_structure.md
antonnikitin97: + friendlies.

PeterJCLaw: There's also no mention of the Challenges either. I deliberately didn't mention them -- they felt "small" enough not to worry about as well as being a bit more tied to the specifics of the content of the year rather than the structure. The Virtual Competition perhaps is different though :thinking: I don't think the friendlies are though (plus we don't always run them).

robots_101/team_supervisor.md
PeterJCLaw: I'd prefer not to -- I think it would be quite hard for one, and probably distract from what we're saying. I had however assumed that it would be a well understood phrase in the general context of things organised for children. Do you think that anyone reading this is likely to not understand what this means?

robots_101/programme_structure.md
antonnikitin97: Agreed with the challenges and friendlies - may be worth mentioning the virtual comp.

robots_101/team_supervisor.md
antonnikitin97: Yeah, I think you're right. Looking at it again I can't really see a way someone who wants to supervise a team wouldn't know what that term meant...

robots_101/programme_structure.md
PeterJCLaw: 30a041a

robots_101/programme_structure.md
```diff
@@ -0,0 +1,100 @@
...
+---
+
+# Robots 101 - Programme Structure
+
+Student Robotics is an annual robotics competition for people aged 16-19.
+But what actually happens during the course of the competition year?
+
+## Registration
+
+Registration for the competition usually opens in late summer. Teams usually
```
sedders123, line 13: We try not to do things during summer holiday for teams
```suggestion
Registration for the competition usually opens in early September. Teams usually
```

sedders123 (approved):

robots_101/programme_structure.md
PeterJCLaw: Hrm, I had actually thought we'd opened it earlier than this quite a few times but I can't find many examples. We do unfortunately end up opening in October quite a bit looking at the team-emails, though as we don't want to be doing that probably best not to advertise that.

PR #612, opened 2024-09-03
Add documentation for kit return

RealOrangeOne:
We often send teams details of kit return when they drop out, but it's probably a good idea to have it generally available.

This information was derived from previous competitions, and the [Kit Shipping](https://studentrobotics.org/runbook/kit/logistics/transport/) pages of the runbook.

### DIFF
--- a/kit/return.md
+++ b/kit/return.md
@@ -15,7 +15,7 @@ Be sure to take all of your own kit with you - it may not be possible to return 
 
 ## Kit Shipping
 
-If you are unable to return your kit at the competition, it must be shipped to us at <a href="{{ site.url }}/contact/">this address</a>, unless otherwise instructed.
+If you are unable to return your kit at the competition, it must be shipped to us at [this address]({{ site.url }}/contact/), unless otherwise instructed.
 
 The Kit List document included with your kit has the date by which the kit needs to be returned.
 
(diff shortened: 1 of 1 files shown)


raccube (approved):

kit/return.md
```diff
@@ -0,0 +1,40 @@
...
+
+After the competition event, you will need to extract the kit from your robot and return them. If you are not using every component of the kit, it should still be brought to the competition so it can be returned. Your returned kit will be checked as you leave the venue. If you have forgotten any items, they must be returned as soon as possible after the event.
+
+<div class="info">
+Be sure to take all of your own kit with you - it may not be possible to return forgotten items to you promptly.
+</div>
+
+## Kit Shipping
+
+If you are unable to return your kit at the competition, it must be shipped to us at <a href="{{ site.url }}/contact/">this address</a>, unless otherwise instructed.
```
PeterJCLaw, line 18:
```suggestion
If you are unable to return your kit at the competition, it must be shipped to us at [this address]({{ site.url }}/contact/), unless otherwise instructed.
```

PeterJCLaw: However I'm not sure if that url is right? If this is meant to point to https://studentrobotics.org/ we prefer to make that explicit rather than using `{{ site.url }}`.

RealOrangeOne: I copied it from the footer links, which use the same code.

PR #617, opened 2024-10-04
Add general information about our events

PeterJCLaw:
This mostly serves as a place to document that we don't require the primary contact to be the person who brings a team to our events, which is something that we're asked every year. We've pretty much always had the same policy, however tend to need to re-evaluate it every time it's asked as we haven't written it down.

Fixes https://github.com/srobo/tasks/issues/1475

### DIFF
--- a/robots_101/events.md
+++ b/robots_101/events.md
@@ -1,0 +1,40 @@
+---
+layout: page
+title: Robots 101 - Events
+---
+
+# Robots 101 - Events
+
+There are three types of event which are held through the [competition year][structure]:
+
+* [Kickstart][kickstart]
+* [Tech Days][tech-days]
+* [Competition][competition]
+
+All events are listed in the [events calendar][events-calendar] on our website.
+The event page for each one contain details of when & where they are, along with
+anything specific you'll need for the day. Generally you will be provided with
+power and internet access, as well as a space to work on your robot, though each
+event is different so be sure to check the event pages.
+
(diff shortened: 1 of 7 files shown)


PeterJCLaw:
Something which does occur to me here is that we're now slightly pushing the definition of the "Robots 101" section, however we don't (yet) have a better place to put this sort of thing. We should probably come back to this and create a section aimed more at supervisors where we can put this more organisational information, separate from the info about robot building.

PeterJCLaw:
Added @andybusse and @Scarzy for trustee visibility. I think this is in line with our existing policies, but do shout if not.

robots_101/events.md
```diff
@@ -0,0 +1,40 @@
...
+
+All events are listed in the [events calendar][events-calendar] on our website.
+The event page for each one contain details of when & where they are, along with
+anything specific you'll need for the day. Generally you will be provided with
+power and internet access, as well as a space to work on your robot, though each
+event is different so be sure to check the event pages.
+
+All teams attending an event **must** be accompanied by a responsible adult.
+This person will typically be the primary Team Supervisor, however does not need
+to be. In cases where this role is delegated, we require only that:
```
PeterJCLaw, line 22: @Scarzy noted that we should not say "only" here.

PeterJCLaw: 3bf51d0414548215ea518c411bcb42392217e8f4

Alexbruvv (approved):

PR #644, opened 2024-11-23
Add robots 101 page about motors

RealOrangeOne:


robots_101/motors.md
```diff
@@ -0,0 +1,112 @@
+---
+layout: page
+title: Robots 101 - Motors
```
RealOrangeOne, line 3: Thought: "Motors" in "Robots 101" could imply there's more here than just how to use the motors API.

Perhaps putting this content under "Robots 101" isn't the right place?

PeterJCLaw: Definitely. From the title I'd assumed this was also about motors themselves, perhaps including suggestions on how to pick them.

PeterJCLaw: Looking again this page feels like it has a lot of overlap with https://studentrobotics.org/docs/tutorials/basic_motor_control, is that expected?
There is good/new content in this PR, which would definitely be valuable to have in our docs, so maybe an alternative is to review the existing tutorial and either merge this or break out separate sub-pages?

PR #651, opened 2024-12-10
Improve syntax highlighting

### LINKED ISSUE #530: Update theme for inline code and code blocks
The current theme makes the inline code blobs looks very similar to regular text and certain highlighting has poor contrast.

image

image

RealOrangeOne:
Fixes #530 

This PR makes 2 main changes:

- Dynamically generate the syntax highlighting styles, so it's clear where they came from.
- Use a better theme with better contrast
- Add a background to inline highlight elements so they're more distinct from the rest of the text

<details><summary>Images</summary>
<p>

image

image

</p>
</details>


_includes/head.html
```diff
@@ -17,6 +17,7 @@
 
   <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/font-awesome/6.2.0/css/all.min.css" integrity="sha512-xh6O/CkQoPOWDdYTDqeRdPCVd1SpvCA9XXcUnZS2FmJNp1coAFzvtCN9BmamE+4aHK8yyUHUSCcJHgXloTyT2A==" crossorigin="anonymous" referrerpolicy="no-referrer" />
   <link rel="stylesheet" href="{{ '/css/main.css' | prepend: site.baseurl }}">
+  <link rel="stylesheet" href="{{ '/css/syntax-highlight.css' | prepend: site.baseurl }}">
```
PeterJCLaw, line 20: This doesn't feel optimal in terms of content delivery -- can we stick with the file being compiled into the `main.css`?

Rakefile
```diff
@@ -34,11 +34,16 @@ end
 
 task :submodules => ['_sass/brand/.git']
 
-task :dev => [:dependencies, :submodules] do
+task :styles => [:dependencies] do
+  highlight_css = `bundle exec rougify style github`.strip()
+  File.write("_sass/_syntax-highlighting.scss", highlight_css)
+end
```
PeterJCLaw, line 40: > Dynamically generate the syntax highlighting styles, so it's clear where they came from.

I'm not sure this makes it any clearer where the styles come from? Having a comment in the file or something might be an alternative (and would avoid additional compile-time overhead). Even if we also provide the command which generates the file and validate in CI that the file matches that command (the [validate-generated-files](https://github.com/marketplace/actions/validate-generated-files) action can help with the latter)

PeterJCLaw (commented):
Some thoughts with an infra-y hat on. (Note also this appears to be broken in CI)

Rakefile
RealOrangeOne: The question is, where did the current file come from? It's hard to tell. No need for that if it's generated.

PeterJCLaw: Somewhat, though a comment at the top of the file like:
```
// This file was generated by running ....
```
Feels like it'd be at least as good?

RealOrangeOne: Yes, for the sake of source attribution alone, that's functionally the same. But this way, there's no longer a generated file committed into the repository, since it can be easily generated automatically at build time.

_includes/head.html
raccube: How? It's served over h2 so there's no difference

RealOrangeOne: It's not preloaded, so there is still a difference. But a single CSS file isn't really going to kill our performance. With that said, it's not especially difficult to put it back into the main CSS bundle.

WillB97 (approved):
Great improvement, could you what colour is used for comment lines.

RealOrangeOne:
@WillB97 contrast looks pretty good for comments:

image

_includes/head.html
RealOrangeOne: Turns out, it was already in the bundle. This was left over from a previous iteration.

raccube (approved):

PR #653, opened 2024-12-21
Canonicalise around "Virtual League" (rather than "Virtual Competition")

PeterJCLaw:
This is more accurate as well as what we've called it in the rules this year. We should stick to this term in future too.


WillB97:
Can we please follow the proper process of having changes reviewed before merging them.

PR #654, opened 2024-12-21
Document use of the code-submitter

PeterJCLaw:
While it's pretty easy to use, having somewhere that we can point people rather than needing to either include these instructions all over the place (and make them up each time) is very useful.

Builds on https://github.com/srobo/docs/pull/653 to avoid conflicts.

### DIFF
--- a/_data/sidebar_tree.yaml
+++ b/_data/sidebar_tree.yaml
@@ -45,6 +45,8 @@ tree:
       title: Updating your brain board
     - url: /tutorials/discord
       title: How to use Discord
+    - url: /tutorials/code_submitter
+      title: How to Submit Code
   - url: /rules/
     title: Rules
     tree:
--- a/_data/sidebar_tree.yaml
+++ b/_data/sidebar_tree.yaml
@@ -46,7 +46,7 @@ tree:
     - url: /tutorials/discord
       title: How to use Discord
     - url: /tutorials/code_submitter
-      title: How to Submit Code
+      title: Using the Code Submitter
   - url: /rules/
     title: Rules
     tree:
--- a/robots_101/programme_structure.md
+++ b/robots_101/programme_structure.md
@@ -69,8 +69,8 @@ as volunteers able to help with the kits and hands-on guidance with your robot.
 
 ## Virtual League
 
-The Virtual League is often the first change you'll have to test your
-strategies against other teams. You'll submit your robot code for running
+The Virtual League is often the first change you'll have to test your strategies
+against other teams. You'll [submit][code-submitter] your robot code for running
 against other teams in the [simulator][simulator]. The matches are livestreamed
 on our [YouTube Channel][youtube] and you'll get a copy of your robot's logs
 afterwards so you can continue to improve it.
@@ -100,5 +100,6 @@ afternoon you'll advance into the knockout stages, and can go on to win prizes.
 [kit]: {{ site.baseurl }}/kit/
 [discord]: {{ site.baseurl }}/tutorials/discord
 [tech-days]: {{ site.baseurl }}/robots_101/tech_days
+[code-submitter]: {{ site.baseurl }}/tutorials/code_submitter
 [simulator]: {{ site.baseurl }}/simulator/
 [youtube]: https://www.youtube.com/@studentrobotics
--- a/robots_101/programme_structure.md
+++ b/robots_101/programme_structure.md
@@ -69,7 +69,7 @@ as volunteers able to help with the kits and hands-on guidance with your robot.
 
 ## Virtual League
 
-The Virtual League is often the first change you'll have to test your strategies
+The Virtual League is often the first chance you'll have to test your strategies
 against other teams. You'll [submit][code-submitter] your robot code for running
 against other teams in the [simulator][simulator]. The matches are livestreamed
 on our [YouTube Channel][youtube] and you'll get a copy of your robot's logs
--- a/tutorials/code_submitter.md
+++ b/tutorials/code_submitter.md
@@ -1,9 +1,9 @@
 ---
 layout: page
-title: How to Submit Code
+title: Using the Code Submitter
 ---
 
-# How to Submit Code
+# Using the Code Submitter
 
 For virtual aspects of the competition, such as the [Virtual League][virtual-league], you will submit your code using the [code-submitter][code-submitter].
 Your code will then be run alongside other teams' code in the [simulator][simulator].
--- a/tutorials/code_submitter.md
+++ b/tutorials/code_submitter.md
@@ -1,0 +1,27 @@
+---
+layout: page
+title: How to Submit Code
+---
+
+# How to Submit Code
+
+For virtual aspects of the competition, such as the [Virtual League][virtual-league], you will submit your code using the [code-submitter][code-submitter].
+Your code will then be run alongside other teams' code in the [simulator][simulator].
+
+To prepare your code for upload, you should create a zip archive containing all the code you want to submit.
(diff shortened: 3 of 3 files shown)


tutorials/code_submitter.md
```diff
@@ -0,0 +1,27 @@
...
+1. Log in using:
+  * your team's TLA as the username, in UPPERCASE\
+    (your team's channel in Discord is named like `#team-tla`)
+  * the same password as you used to get access to your team's area in Discord\
+    (check with your supervisor if you're not sure what this is)
+1. Upload the archive to the Code Submitter
+
+[virtual-league]: {{ site.baseurl }}/robots_101/programme_structure#virtual-league
+[simulator]: {{ site.baseurl }}/simulator/
+[code-submitter]: https://studentrobotics.org/code-submitter/
```
ShadowLord2005, line 27: Should this use the {{ site.baseurl }} format as well?

tutorials/code_submitter.md
```diff
@@ -0,0 +1,27 @@
...
+For virtual aspects of the competition, such as the [Virtual League][virtual-league], you will submit your code using the [code-submitter][code-submitter].
+Your code will then be run alongside other teams' code in the [simulator][simulator].
+
+To prepare your code for upload, you should create a zip archive containing all the code you want to submit.
+As with running the code in the simulator locally or on your robot, this must include a file named _exactly_ **`robot.py`**, placed in the root of the archive (not within a subfolder).
+You can also provide other files as well and use them from your `robot.py` file, however the code-submitter will always look for a `robot.py` file as that is what the simulator will run.
+
+Once you have prepared the archive containing your code:
+
+1. Visit the [Code Submitter][code-submitter]
```
ShadowLord2005, line 17: Don't think all these numbers should be 1.

tutorials/code_submitter.md
```diff
@@ -0,0 +1,27 @@
...
+layout: page
+title: How to Submit Code
+---
+
+# How to Submit Code
+
+For virtual aspects of the competition, such as the [Virtual League][virtual-league], you will submit your code using the [code-submitter][code-submitter].
+Your code will then be run alongside other teams' code in the [simulator][simulator].
+
+To prepare your code for upload, you should create a zip archive containing all the code you want to submit.
```
samjmartin04, line 11: Do we want to give any information about how to make a zip archive? Even if that's something like "on windows, you can do this by selecting all the files you wish to include, right clicking, and selecting archive files"

robots_101/programme_structure.md
```diff
@@ -69,8 +69,8 @@ as volunteers able to help with the kits and hands-on guidance with your robot.
 
 ## Virtual League
 
-The Virtual League is often the first change you'll have to test your
-strategies against other teams. You'll submit your robot code for running
+The Virtual League is often the first change you'll have to test your strategies
```
samjmartin04, line 72:
```suggestion
The Virtual League is often the first chance you'll have to test your strategies
```

tutorials/code_submitter.md
PeterJCLaw: I can see why you'd ask that, but no. `{{ site.basurl }}` varies with local vs production hosting to account for the different hosting setups. So we use it for internal links which need that but not for external ones, which need to be absolute so that they work consistently. (Once you know this it also becomes a handy way to know which are which and thus how they should be verified)

tutorials/code_submitter.md
ShadowLord2005: What level of competency do we assume? It might be good however I assume that we would then want to also include guides for Linux and MacOS?

tutorials/code_submitter.md
PeterJCLaw: I can see why you'd ask, however this is a deliberate markdown trick to avoid redundant diff noise if a new item needs adding in the middle of the list. It'll still render correctly. The numeric markers just mean "this is a numbered list" -- the individual numbers used are ignored by the rendering logic in either case.

tutorials/code_submitter.md
ShadowLord2005: Ok, thanks for clarifying

tutorials/code_submitter.md
PeterJCLaw: I thought about this, but opted not to for simplicity/speed ... and intended to leave a comment about it but then forgot. Very open to that being added, but felt this page was valuable enough without it and I'm not well placed right now to chase down exactly how to do this across platforms.

tutorials/code_submitter.md
ShadowLord2005: Ah, ok. I'm not familiar with all markdown tricks.

tutorials/code_submitter.md
PeterJCLaw: We typically provide info for Windows & MacOS. I think omitting Linux is _probably_ ok (as Linux feels likely to only be personal devices and the plethora of desktop environments makes it harder anyway), though I do think we should start including Chromebooks (see https://github.com/srobo/docs/issues/655).

ShadowLord2005: Sounds good. More info can't hurt.

PeterJCLaw: For clarity: I am still proposing that we aim to ship this PR without any platform walkthroughs and that we can add that later if/when needed.

ShadowLord2005: I'm happy with that.

raccube: There are instructions for this here: https://studentrobotics.org/docs/tutorials/getting_code_on_the_robot

_data/sidebar_tree.yaml
```diff
@@ -45,6 +45,8 @@ tree:
       title: Updating your brain board
     - url: /tutorials/discord
       title: How to use Discord
+    - url: /tutorials/code_submitter
+      title: How to Submit Code
```
RealOrangeOne, line 49: Nit: This could be misread as how code is submitted all the time, rather than just in a virtual context. Perhaps "Using the Code Submitter"?

tutorials/code_submitter.md
ShadowLord2005: That isn't quite what we're after as we specifically are talking about instructions for creating a `.zip` folder

_data/sidebar_tree.yaml
PeterJCLaw: Thanks, yeah wasn't entirely happy with this title. Finding something which doesn't wrap is a challenge though!

PeterJCLaw: That fits! Thanks :+1: https://github.com/srobo/docs/pull/654/commits/f7b55159fe43f030e8caf4a92bffd84801fcb7d9

robots_101/programme_structure.md
```diff
@@ -69,8 +69,8 @@ as volunteers able to help with the kits and hands-on guidance with your robot.
 
 ## Virtual League
 
-The Virtual League is often the first change you'll have to test your
-strategies against other teams. You'll submit your robot code for running
+The Virtual League is often the first chance you'll have to test your strategies
+against other teams. You'll [submit][code-submitter] your robot code for running
 against other teams in the [simulator][simulator]. The matches are livestreamed
 on our [YouTube Channel][youtube] and you'll get a copy of your robot's logs
```
antonnikitin97, line 75: > and you'll get a copy of your robot's logs

Pedantry: Don't they also get their match recording as well?

PeterJCLaw: Yeah, probably. Feel free to open a PR for that. I'd rather keep this one focussed on the code submitter.

samjmartin04 (approved):

PR #661, opened 2025-01-30
More detail on what the Virtual League is

PeterJCLaw:
This aims to better help rookies, be a target for related links and a place we can refer to for how to describe the virtual league in other media.

See also https://github.com/srobo/website/pull/599

### DIFF
--- a/robots_101/programme_structure.md
+++ b/robots_101/programme_structure.md
@@ -78,7 +78,7 @@ array of sensors.
 This allows you to use the simulator and the Virtual League to learn about
 writing code for a robot as well as developing the code for your physical robot
 before it's fully built. The Virtual League is thus a both a first test of your
-code as well as an early opportunity to explore strategies competing against
+code as well as an early opportunity to explore strategies for competing against
 other teams.
 
 Shortly before the event you'll [submit][code-submitter] so the matches can be
--- a/robots_101/programme_structure.md
+++ b/robots_101/programme_structure.md
@@ -70,10 +70,10 @@ as volunteers able to help with the kits and hands-on guidance with your robot.
 ## Virtual League
 
 The Virtual League is often the first chance you'll have to test your strategies
-against other teams. This uses the same [simulator][simulator] used for the
-pre-Kickstart activities and the includes a [simulated robot][simulated-robot]
-which both supports the same programming interface as your robot and has a wide
-array of sensors.
+against other teams. The Virtual League plays the same game as the physical
+competition, though takes place in the [simulator][simulator]. You'll write
+code for a [simulated robot][simulated-robot] which both supports the same
+programming interface as your physical robot and has a wide array of sensors.
 
 This allows you to use the simulator and the Virtual League to learn about
 writing code for a robot as well as developing the code for your physical robot
@@ -81,10 +81,10 @@ before it's fully built. The Virtual League is thus a both a first test of your
 code as well as an early opportunity to explore strategies for competing against
 other teams.
 
-Shortly before the event you'll [submit][code-submitter] so the matches can be
-simulated. The matches are then livestreamed on our [YouTube Channel][youtube].
-After the livestream you'll get a copy of your robot's logs afterwards so you
-can continue to improve it.
+Shortly before the event you'll [submit][code-submitter] your robot code so the
+matches can be simulated. The matches are then livestreamed on our
+[YouTube Channel][youtube]. Afterwards you'll get a copy of your simulated
+robot's logs so you can continue to improve your code.
 
 ## Competition
 
--- a/robots_101/programme_structure.md
+++ b/robots_101/programme_structure.md
@@ -70,10 +70,21 @@ as volunteers able to help with the kits and hands-on guidance with your robot.
 ## Virtual League
 
 The Virtual League is often the first chance you'll have to test your strategies
-against other teams. You'll [submit][code-submitter] your robot code for running
-against other teams in the [simulator][simulator]. The matches are livestreamed
-on our [YouTube Channel][youtube] and you'll get a copy of your robot's logs
-afterwards so you can continue to improve it.
+against other teams. This uses the same [simulator][simulator] used for the
+pre-Kickstart activities and the includes a [simulated robot][simulated-robot]
+which both supports the same programming interface as your robot and as a wide
+array of sensors.
+
+This allows you to use the simulator and the Virtual League to learn about
+writing code for a robot as well as developing the code for your physical robot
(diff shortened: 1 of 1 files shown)


robots_101/programme_structure.md
```diff
@@ -70,10 +70,21 @@ as volunteers able to help with the kits and hands-on guidance with your robot.
 ## Virtual League
 
 The Virtual League is often the first chance you'll have to test your strategies
-against other teams. You'll [submit][code-submitter] your robot code for running
-against other teams in the [simulator][simulator]. The matches are livestreamed
-on our [YouTube Channel][youtube] and you'll get a copy of your robot's logs
-afterwards so you can continue to improve it.
+against other teams. This uses the same [simulator][simulator] used for the
+pre-Kickstart activities and the includes a [simulated robot][simulated-robot]
+which both supports the same programming interface as your robot and as a wide
```
ShadowLord2005, line 75:
```suggestion
which both supports the same programming interface as your robot and has a wide
```

robots_101/programme_structure.md
```diff
@@ -70,10 +70,21 @@ as volunteers able to help with the kits and hands-on guidance with your robot.
...
+array of sensors.
+
+This allows you to use the simulator and the Virtual League to learn about
+writing code for a robot as well as developing the code for your physical robot
+before it's fully built. The Virtual League is thus a both a first test of your
+code as well as an early opportunity to explore strategies competing against
+other teams.
+
+Shortly before the event you'll [submit][code-submitter] so the matches can be
```
ShadowLord2005, line 84:
```suggestion
Shortly before the event you'll [submit][code-submitter] your code so the matches can be
```
Can you check this doesn't interfere with adding the link

robots_101/programme_structure.md
```diff
@@ -70,10 +70,21 @@ as volunteers able to help with the kits and hands-on guidance with your robot.
...
+simulated. The matches are then livestreamed on our [YouTube Channel][youtube].
+After the livestream you'll get a copy of your robot's logs afterwards so you
```
ShadowLord2005, line 86: Don't think you need need both after and afterwards. Not sure which is better

robots_101/programme_structure.md
```diff
@@ -70,10 +70,21 @@ as volunteers able to help with the kits and hands-on guidance with your robot.
...
+can continue to improve it.
```
ShadowLord2005, line 87: Do we need to be more specific with it?

robots_101/programme_structure.md
ShadowLord2005: Is it worth explicitly mentioning physical to really make the point?
```suggestion
which both supports the same programming interface as your physical robot and as a wide
```
Not sure if adding the suggestions is useful for your or just putting the thoughts?

robots_101/programme_structure.md
```diff
@@ -70,10 +70,21 @@ as volunteers able to help with the kits and hands-on guidance with your robot.
 ## Virtual League
 
 The Virtual League is often the first chance you'll have to test your strategies
-against other teams. You'll [submit][code-submitter] your robot code for running
-against other teams in the [simulator][simulator]. The matches are livestreamed
-on our [YouTube Channel][youtube] and you'll get a copy of your robot's logs
-afterwards so you can continue to improve it.
+against other teams. This uses the same [simulator][simulator] used for the
+pre-Kickstart activities and the includes a [simulated robot][simulated-robot]
```
WillB97, line 74: It's not the same simulator as the pre-kickstart activities. It's the version released at kickstart.

PeterJCLaw: Hrm, technically it uses the same _simulator_ though it may not use the _simulated world_. This could probably be clarified though.

robots_101/programme_structure.md
PeterJCLaw: What do you have in mind?

In my head the expected audience here is "people who want to know roughly what the competition structure is", rather than "people trying to actually submit an entry for the Virtual League". (Though I have linked onwards to things the latter group hopefully would find useful).

robots_101/programme_structure.md
PeterJCLaw: > Not sure if adding the suggestions is useful for your or just putting the thoughts?

Both/either. Suggestions can be useful to clarify what you're (as a reviewed) suggesting even if the author choses not to use them directly. I've even (very rarely) used them when (as a reviewer) I'm not intending for them to be used directly, though I'd preface such cases with "something like" or "untested" to indicate that.

robots_101/programme_structure.md
ShadowLord2005: I think something like "your robot" or "your code". To me I think it could be confused with improving your simulated robot though possibly not if I consider your context more.

PeterJCLaw: Oh, I see, you mean the final "it". Aha, yeah, explicitly saying "your robot code" here sounds like a good idea.

robots_101/programme_structure.md
PeterJCLaw: 561cbf4e7458685fd3f30e1b50afa520a5638d7f

robots_101/programme_structure.md
PeterJCLaw: 561cbf4e7458685fd3f30e1b50afa520a5638d7f

robots_101/programme_structure.md
PeterJCLaw: 561cbf4e7458685fd3f30e1b50afa520a5638d7f

robots_101/programme_structure.md
PeterJCLaw: 561cbf4e7458685fd3f30e1b50afa520a5638d7f, though not sure what you mean by "adding the link"?

robots_101/programme_structure.md
PeterJCLaw: 561cbf4e7458685fd3f30e1b50afa520a5638d7f

robots_101/programme_structure.md
ShadowLord2005: I presume the `[submit][code-submitter]` is some sort of internal link, being unsure how it works I wanted you to check that what I added didn't interfere with it.

PeterJCLaw: Oh, right. That's a markdown link but one which uses a footnote to express what the link target is. For example the following are equivalent:
```markdown
- [SR](https://srobo.org/)
- [SR][srobo]

[srobo]: https://srobo.org/
```

ShadowLord2005 (approved):
All looks good to me, happy with all the changes

PR #665, opened 2025-02-06
Document `robot.sleep` and `robot.time`

RealOrangeOne:
These methods exist, and are incredibly useful, but aren't documented.

### DIFF
--- a/programming/robot_api/index.md
+++ b/programming/robot_api/index.md
@@ -128,9 +128,9 @@ sleep(seconds)
     See [Simulation of Time]({{ site.baseurl }}/simulator/using_the_simulator#simulation-of-time) for more information.
 
 time()
-:   Returns the current time in seconds, measured since the [epoch](https://docs.python.org/3/library/time.html#epoch) (January 1, 1970, midnight). This method is similar to the built-in [`time.time`](https://docs.python.org/3/library/time.html#time.time) method.
+:   Returns the current time in seconds, measured since an [epoch](https://en.wikipedia.org/wiki/Epoch) (reference time). This method is similar to the built-in [`time.time`](https://docs.python.org/3/library/time.html#time.time) method.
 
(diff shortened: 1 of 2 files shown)


programming/robot_api/index.md
```diff
@@ -108,6 +108,43 @@ is_simulated
...
+    ~~~~~
+
+    This method is particularly useful in the simulator, where the simulation may be running faster than real-time. Whilst `time.sleep` will still work as expected on the physical robot, it's still recommended to use `robot.sleep` to ensure your code is portable.
+
+    See [Simulation of Time]({{ site.baseurl }}/simulator/using_the_simulator#simulation-of-time) for more information.
+
+time()
+:   Returns the current time in seconds, measured since the [epoch](https://docs.python.org/3/library/time.html#epoch) (January 1, 1970, midnight). This method is similar to the built-in [`time.time`](https://docs.python.org/3/library/time.html#time.time) method.
+
+    The exact time on the robot will not be correct, however it can still be used to measure the duration between 2 events.
```
PeterJCLaw, line 133: Observation: given that we know that this value will never actually be right, it feels a little bit odd to frame the epoch as a fixed point.

:bulb: suggestion: maybe re-frame this along the lines of:
```suggestion
:   Returns the current time in seconds, measured since an [epoch](https://en.wikipedia.org/wiki/Epoch) (reference time). This method is similar to the built-in [`time.time`](https://docs.python.org/3/library/time.html#time.time) method.

    As your robot is unable to keep time when it is turned off the epoch used is different each time the robot is powered on. While powered on, the epoch will not change.

    This means that while the exact time on the robot will not be correct (in the sense that it won't match your ~watch~ phone), it can still be used to measure the duration between two points in time
```

PeterJCLaw (approved):
:+1: some suggestions for enhancements

PeterJCLaw (approved):

PR #669, opened 2025-03-19
Use a better times symbol

PeterJCLaw:
Letter 'x' is not a times symbol.


RealOrangeOne:
Is this all occurrences of an "X" used as a times symbol? Seems silly to change just one.

PeterJCLaw:
The others were dimensions, have done that in the second commit.

This turns out to be handy.
```
git grep -P '\bx\s*\d+' -- :^images :^package-lock.json :^resources
```
The `:^` syntax excludes those files.

raccube (approved):

